Issue #1, opened 2019-10-16
Namespace for modules


certik:
So you can require module data to be referenced like module.var:
```fortran
use utils, only
...
call utils%savetxt()
```
One issue with just `use utils` is that it still pollutes your local namespace with `savetxt`, so that's the reason for `use utils, only` (perhaps one could do `use utils, only:`).

Another alternative is `use namespace utils` (suggested by [Milan Curcic](https://twitter.com/realmilancurcic/status/1184269876543799296)). Or perhaps `use, namespace :: utils`, to by compatible with the existing syntax like `use, intrinsic :: iso_fortran_env`.

Initially discussed at https://twitter.com/OndrejCertik/status/1184245450976575488.

certik:
The paper is submitted at https://j3-fortran.org/doc/year/19/19-246.txt. I will try to get feedback on it from the committee.

septcolor:
+1 for this proposal because, in addition to that many other languages
provide this feature, I often create a dummy derived type (having a lot of
NOPASS module variables) to mimic this behavior (which is very tedious).

Indeed, this seems to be almost a "frequently asked request (FAR)" (not FAQ :-)
http://fortranwiki.org/fortran/show/Extension+to+USE+statement

cmacmackin:
Agree, this seems like a good feature.

certik:
So the committee unfortunately did not consider this proposal at the meeting last week. I was hoping to get at least a general "yes" or "no" direction from the committee. But I talked to a few members directly, and I did not hear any "this is a bad idea", or "this should not be done". 

Here is my plan:

1. Write a high quality full proposal for this. This means more use cases, we have to brainstorm any possible issues, and some members of the committee like to structure any new features as "requirements", "specification" and "syntax", so we should incorporate all that.

2. I will ask on the j3 mailinglist once a more complete proposal is ready to get some feedback

3. I will ensure that this gets to the committee program for next meeting to be official considered for the 202y standard.

septcolor:
I fully agree with plan (1-3), but I have one concern (which may apply to other proposals as well). Though I may be too nervous, I will explain it anyway...

[...]

PR #8, opened 2019-10-16
Initial proposal for namespace for modules

certik:
Work on #1.

@zingale, I talked about this with a few committee members, and they recommended to start collecting use cases first, then see what the reaction of the committee is, then go from there. So this is my initial draft, that I would like to submit tomorrow (the committee meeting is till this Friday). Do you have some use cases that I didn't mention in the proposal?


zingale:
I think that looks great.  Both avoiding name clashes and making it easier to understand where a function / data is coming from are my motivation for wanting namespaces.

PR #31, opened 2019-10-21
Add proposal for protected attribute in derived types

aradi:


certik:
@aradi, thanks for submitting this. I have a few questions. Wouldn't your example work with current Fortran already? I would like to see an example which does not work with current Fortran, and shows how the "protected" attribute improves things.

certik:
Ah I see --- it's something like `const` in C++? It makes the attribute constant / read only. I thought this was about making the member "private".

aradi:
I added some more explanation. The case presented would be only possible, if the component were made `public`, but that would jeopardize data consistency.

aradi:
Yes, it was meant like `const` in C++ (and not as `protected` in C++ :wink: )

zbeekman:
I've wanted this for a long time. Writing a ton of getter methods is a pain in the ass and a waste of time. An attribute that would only allow modification by TBPs would be really nice, and would allow for direct reading of UDT components.

certik:
@aradi I did not forget about this PR --- I first want to figure out a workflow with the committee how to consider proposals such as this one. Things will start moving in the next few weeks.

aradi:
All fine, thanks in advance and especially thanks for all your efforts into involving the community!

certik:
@aradi, can you comment on #122 and (hopefully) list it among your top 5? We'll now start drafting proposals, so we'll figure out some good template and help you finish it.

aradi:
@zjibben Thanks! I think, the proposal is trivial enough not to have side effects or unwanted implications, this is why there was not much discussion about it.

jvdp1:
Thanks for this proposal. It is something I tried intuitively once (without success of course ;) ), before reading one more time the documentation of `protected`.

certik:
I assigned to @zjibben to merge when he thinks it's ready.

zjibben:
I went ahead and merged. I agree, I think this is straightforward enough to not need more clarification, but if anyone thinks of something please leave a comment or submit a PR! I’ll submit this weekend otherwise. Thanks for the proposal @aradi !

Issue #22, opened 2019-10-18
Allow setting default value for optional dummy argument


milancurcic:
## Problem

Currently, Fortran doesn't allow setting a default value for optional dummy arguments. If declaring a dummy argument as `optional`, the user must:

1. Manually make appropriate checks about whether the actual argument is provided by the caller;
2. Use a separate variable inside the procedure because the code must not reference the optional dummy argument if not present.

## Example

Example of a quadratic function that optionally allows evaluation of its linear form:

```fortran
real function quadratic(x, a, b, c)  
  ! returns a + b * x + c * x**2 if c is present and a + b * x otherwise
  real, intent(in) :: x, a, b
  real, intent(in), optional :: c
  real :: c_tmp ! use separate variable to avoid referencing non-present arg
  if (present(c)) then
    c_tmp = c
  else
    c_tmp = 0 ! default value if c is not present
  end if
  quadratic = a + b * x + c_tmp * x**2
end function quadratic
```
Checking for presence of the optional argument and using a temporary variable is cumbersome and error-prone, and it is even more so if the same needs to be done for many optional arguments.

## Solution

Allow setting the default value of the optional argument in the `function` (or `subroutine`) statement by using the assignment operator `=`:

```fortran
real function quadratic(x, a, b, c=0)
  ! returns a + b * x + c * x**2 if c is present and a + b * x otherwise
  real, intent(in) :: x, a, b
  real, intent(in), optional :: c
  quadratic = a + b * x + c * x**2
end function quadratic
```
This is similar to how Python keyword arguments are defined.

## Comments

* The proposed syntax is complementary to the existing syntax, that is, it doesn't break existing code;
* Its style is more or less Fortranic and consistent with other syntax elements;
* Same syntax as Python keyword args, so it would be intuitive to Python developers learning Fortran;
* Could be implemented completely as a pre-processor, for a proof-of-concept.

certik:
Yes, great idea. I usually do it like this:
```fortran
real function quadratic(x, a, b, c)  
  ! returns a + b * x + c * x**2 if c is present and a + b * x otherwise
  real, intent(in) :: x, a, b
  real, intent(in), optional :: c
  real :: c_
  c_ = 0
  if (present(c)) c_ = c
  quadratic = a + b * x + c_ * x**2
end function quadratic
```
Which is simpler than your original code, but it's still cumbersome to always introduce the local variable and do the `present(c)` check.

[...]

PR #137, opened 2020-01-10
proposal for setting default values of optional arguments

milancurcic:
First draft of the proposal to set default value for optional arguments (#22).

### DIFF
--- a/proposals/default_optional_arguments/proposal.txt
+++ b/proposals/default_optional_arguments/proposal.txt
@@ -91,8 +91,9 @@ dummy argument to have an initializer, which would be triggered only
 when corresponding actual argument is not passed by the caller. 
 
 This improvement has already been suggested by Van Snyder in 
-18-136r1, has received votes in the user survey for 202X (TODO 
-link), and appeared on Data subgroup's wishlist at meeting 215 (see 
+18-136r1, has received votes in the user survey for 202X (see
+https://isotc.iso.org/livelink/livelink?func=ll&objId=19530634&objAction=Open&viewType=1), 
+and appeared on Data subgroup's wishlist at meeting 215 (see 
 https://j3-fortran.org/doc/year/18/18-122r1.txt).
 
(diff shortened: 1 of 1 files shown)


milancurcic:
I will need some help linking to the user survey that @sblionel mentioned (I participated in it, and I read the results back then, but now I can't find it anywhere).

I will also need help suggest a specific and formal addition/change to the standard.

sblionel:
https://isotc.iso.org/livelink/livelink?func=ll&objId=19530634&objAction=Open&viewType=1

jvdp1:
@milancurcic thank you for the proposal. The first draft looks good to me.

> I will also need help suggest a specific and formal addition/change to the standard.

I am not sure what you mean.

Should the following cases be discussed in the draft:

1) What is the behavior of `present`?
```fortran
subroutine foo(dummy)
     integer, intent(in), optional :: dummy = 0
     print*, present(dummy) ! = .true. if dummy provided, .false. otherwise
end subroutine
```
2) How does `dummy` behave when it is `optional` and `allocatable`?
```fortran
subroutine foo(dummy)
     integer, allocatable, intent(in), optional :: dummy(:) = [ 0, 0] 
    if ( .not. present(dummy)) then
     allocate(dummy(5))   !I intentionaly used a dim of 5, instead of 2
     dummy = 1
   end if
end subroutine
```

I think your proposal has no problem when `intent(inout || out), optional` is used since the argument is not provided

sblionel:
@jvdp1 Anything and everything we can think of that seems relevant should be discussed - both of those questions are worthy of mention.

My opinions would be: 

1. `present()` would always return true for an argument with a default.
2. That combination is not valid. If the dummy is allocatable, the actual must also be. 

My position is that a default value has exactly the same meaning as if the caller passed that value as the corresponding actual argument, with all the existing rules applied. Indeed, I would expect a compiler to implement it this way, doing the substitution at the call and not inside the procedure being called.

PR #142, opened 2020-01-28
Traits Proposal

everythingfunctional:
Proposal for traits as a method of enabling more generic programming capabilities.

See #125

### DIFF
--- a/proposals/traits/proposal.txt
+++ b/proposals/traits/proposal.txt
@@ -1,0 +1,75 @@
+To: J3
+From: Brad Richardson
+Subject: Traits for Types
+Date: 2020-January-27
+
+Proposal for Fortran Standard: 202y (NOT 202x)
+
+1. Problem
+
+Currently, in order to pass a derived type to some library expected a type
+derived from it's own abstract type requires the addition of a "wrapper" type
+to be used. This requires a cumbersome amount of boiler plate code and is
+not conducive to the type of "generic" code one would like to be able to write.
+
+This proposal seeks to address the issue that writing "generic" libraries is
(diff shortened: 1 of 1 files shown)


everythingfunctional:
This is just a rough draft. It needs somebody with more experience writing these to help me use the write terms and verbiage, and to make sure we cover all the necessary points.

certik:
@everythingfunctional thanks for staring this!

We discussed all kinds of aspects at the original issue. Let's start with this:

* is this compile time or runtime polymorphism?

everythingfunctional:
So far I haven't specified anything that would force an implementation to use one or the other. I was leaning towards runtime polymorphism though, as it aligns with the current runtime polymorphism already available through inheritance.

I know you were hoping for compile time, but I really think that is better done with templates than traits. I could be persuaded to change it if there is a significant consensus otherwise.

certik:
So traits and templates interact. I've been learning Rust recently, and I like so far how they did it. Although I am still confused what exactly is compile time and runtime --- I'll write up C++ / Rust rosetta stone to clarify that soon. And I think it will be helpful for our discussions here too. So far it seems to me what Rust did is exactly what we should do in Fortran also, but I still need more time to learn.

In order to move things forward --- I think you can do your best clarifying things the way you see fit in the meantime. That will be helpful either way. I will let you know once I made some progress.

certik:
Thanks! I am going to merge it now and upload.

certik:
I uploaded it here:

https://j3-fortran.org/doc/year/20/20-109.txt

(I had to fix up a few things #150.)

PR #145, opened 2020-02-21
Prepare default value for optional arguments proposal for submission.

zjibben:
This continues the default-value saga from #22 #137 and #140.

I made a few formatting changes, added an introduction, and tried to capture some of the discussion in #140 about intent(out) arguments. My goal is to make sure our interests and the possible different behaviors are expressed, so the Committee has some ideas to consider. I'd like to get it in shape to submit before the meeting next week, and hope to discuss it there.

@jvdp1, I added your name to the proposal if you don't mind. Please let me know if you would like this changed and if spelling is correct.

### DIFF
--- a/proposals/default_optional_arguments/proposal.txt
+++ b/proposals/default_optional_arguments/proposal.txt
@@ -166,6 +166,21 @@ intent(out) variable is initially undefined, meaning the logic of
 present() might still be necessary in this case. Furthermore, the behavior
 suggested above where present() = .true. always for variables with a
 specified default would not be desired for intent(out) variables.
+A possible use case for this behavior could be:
+
+    integer function open(filename, mode, iostat) result(u)
+      character(*), intent(in) :: filename
+      character(*), intent(in), optional :: mode
+      integer, intent(out), optional :: iostat = 0
+      character(3) :: mode_
+      character(:),allocatable :: action_, position_, status_, access_, form_
+     !some code
+     !....
+     open(newunit=u, file=filename, &
+         action = action_, position = position_, status = status_, &
+         access = access_, form = form_, &
+         iostat = iostat)
+end function
 
 A third option is that optional intent(out) variables with a specified
 default are always initialized to that default, even when present,
--- a/proposals/default_optional_arguments/proposal.txt
+++ b/proposals/default_optional_arguments/proposal.txt
@@ -163,7 +163,7 @@ intent(out) variables would be usable regardless of whether the value
 is actually passed back to the caller, removing the need for present()
 and "temporary" variables. The complication is that a present
 intent(out) variable is initially undefined, meaning the logic of
-present() is still necessary in this case. Furthermore, the behavior
+present() might still be necessary in this case. Furthermore, the behavior
 suggested above where present() = .true. always for variables with a
 specified default would not be desired for intent(out) variables.
 
--- a/proposals/default_optional_arguments/proposal.txt
+++ b/proposals/default_optional_arguments/proposal.txt
@@ -1,28 +1,32 @@
 To: J3                                                     J3/XX-XXX
-From: Milan Curcic
+From: Milan Curcic, Jeremie Vandenplas, Zach Jibben
 Subject: Default value for optional arguments
 Date: 2020-January-10
-
-Proposal for Fortran Standard: 202y (NOT 202x)
-
-
-Prior art: https://j3-fortran.org/doc/year/18/18-136r1.txt
+Reference: 18-136r1, 18-122r1
 
 
-1. Problem
+1. Introduction
 
-Currently, standard Fortran does not allow setting a default value
-for optional arguments. Default value of an optional argument is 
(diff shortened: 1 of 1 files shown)


proposals/default_optional_arguments/proposal.txt
```diff
@@ -141,8 +142,36 @@ A use case of this behavior of the function present() could be:
...
+
+The benefit is that, just like proposed for intent(in) variables,
+intent(out) variables would be usable regardless of whether the value
+is actually passed back to the caller, removing the need for present()
+and "temporary" variables. The complication is that a present
+intent(out) variable is initially undefined, meaning the logic of
+present() is still necessary in this case. Furthermore, the behavior
+suggested above where present() = .true. always for variables with a
+specified default would not be desired for intent(out) variables.
+
```
jvdp1:
```suggestion
A possible use case for this behaviorcould be:

    integer function open(filename, mode, iostat) result(u)
      character(*), intent(in) :: filename
      character(*), intent(in), optional :: mode
      integer, intent(out), optional :: iostat = 0
      character(3) :: mode_
      character(:),allocatable :: action_, position_, status_, access_, form_
     !some code
     !....
     open(newunit=u, file=filename, &
         action = action_, position = position_, status = status_, &
         access = access_, form = form_, &
         iostat = iostat)
end function

```

This code is actually extracted from [stlib](https://github.com/fortran-lang/stdlib/blob/1e0e460c78dec6e926b51d29fa5db889d5592f7a/src/stdlib_experimental_io.fypp#L233)

proposals/default_optional_arguments/proposal.txt
```diff
@@ -141,8 +142,36 @@ A use case of this behavior of the function present() could be:
...
+A third option is that optional intent(out) variables with a specified
+default are always initialized to that default, even when present,
+rather than leaving them undefined.
```
jvdp1: I am in favor of such a behavior. However, what would the behavior be for:

```fortran
subroutine foo(a)
  real, intent(out), allocatable, optional :: a(:, :) = 0
end subroutine foo
```

proposals/default_optional_arguments/proposal.txt
```diff
@@ -141,8 +142,36 @@ A use case of this behavior of the function present() could be:
...
```
jvdp1: The scenario `intent(inout)` (or no specification of `intent()`) is missing. Its behavior would probably depend on the behaviors of `intent(in)` and `intent(out)`, e.g., as proposed by @everythingfunctional (which might not be in agreement with all different proposed options for `intent(out)`):

"If it is intent(inout) and the caller provides an actual argument, it wil get the value of the actual argument, AND the returned value is equal to the actual value if it is not modified, or the assigned value otherwise. (Use case: @everythingfunctional 's example)."

Should it be mentioned?

proposals/default_optional_arguments/proposal.txt
milancurcic: "behaviorcould" -> "behavior could"

certik: I modified @jvdp1's comment to fix the spelling error and committed.

proposals/default_optional_arguments/proposal.txt
zjibben: This is a very good question, and I think applies to `intent(in)` and `intent(inout)` arguments too (if no actual argument is given). I’ll add a section 5.3 on allocatable arrays, it definitely should be addressed. Off the top of my head, one option is of course to forbid defaults for allocatables. This isn’t preferable, but I suspect allowing defaults for scalars only would still address the majority of use cases. Another is to forbid array broadcasting as a default specifier. This would mean `a(:) = [0,0]` would be permissible, but not `a(:) = 0`. Any other ideas?

proposals/default_optional_arguments/proposal.txt
```diff
@@ -1,28 +1,32 @@
 To: J3                                                     J3/XX-XXX
-From: Milan Curcic
+From: Milan Curcic, Jeremie Vandenplas, Zach Jibben
```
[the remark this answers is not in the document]
zjibben: Thanks!

proposals/default_optional_arguments/proposal.txt
jvdp1: It is indeed also true for `intent(in)` and `intent(inout)`. I have no knowlegde about compiler development, so I have no idea what would be possible to implement or not. The easiest thing is probably to forbid default values for allocatable arrays, but it could complicate the standard.

certik:
It seems there is general agreement on these changes. Further discussion might be needed on top, so let's open issues or further PRs as needed. I merged this, so that we have something in the repository that we can at least roughly agree on, and we can now work on polishing it up so that @zjibben can submit it as an official proposal.

In general, if we can't reach agreement, I think it is perfectly ok to simply propose the alternatives (with pros and cons) in the proposal, and let the committee discuss it and provide feedback.

PR #147, opened 2020-02-22
Add sections on intent(inout) and arrays to optional-default proposal

zjibben:
More on the proposal for #22. @jvdp1 @certik @milancurcic.


zjibben:
Thank you all! I will submit this for our meeting.

PR #149, opened 2020-02-23
Namespaces

certik:
Resubmitting the namespaces proposal


certik:
Uploaded at https://j3-fortran.org/doc/year/20/20-108.txt.

PR #154, opened 2020-02-24
Examples added to namelist delim + reformat

marshallward:
An example section was added to the Namelist delimiter proposal in order
to highlight the problem of non-delimited character array namelist
output.

The text was also reformatted to 72 columns, since this seems to be the
standard used by other proposals.

Other aberrant use of reST syntax was also removed.


certik:
@zjibben would you be willing to sponsor this proposal and upload it?

zjibben:
Thank you @marshallward ! I gave the JOR subgroup in J3 a copy of your paper.

marshallward:
Great!  Thanks very much for handling this for me.

PR #159, opened 2020-02-28
Namelist delimiter proposal update

marshallward:
This patch contains two main modifications to the namelist delimiter
proposal:

1. The introduction has been modified for clarity and to include the
   proposed change, rather than simply a discussion of the problem.

2. The introduction and proposal sections have been modified to clarify
   that list-directed output would be unaffected by the change.


marshallward:
I am not sure how to submit this, and other feedback suggests that I am the only person on the planet who is concerned about this, so I am going to close this.

PR #175, opened 2020-07-12
Revised version of the default arguments proposal

wclodius2:
I have written a revision of the paper by @milancurcic, @jvdp1, and @zjibben that attempts to address most of the issues raised at the Feb. 25 meeting, and a few other issues that have come to my mind. The paper differs from the original primarily by:

Restricting optional arguments to arguments with the INTENT(IN), VALUE, or no intent attributes.

Requiring arguments with a default assignment to be explicitly given the DEFAULT attribute.

Forbidding arguments from having both the DEFAULT and OPTIONAL attributes so that the PRESENT function retains its semantics.

Allowing the expression on the right of the default assignment be a restricted expression and not just a constant expression.

Strongly encouraging that the default assignment be in the FUNCTION or SUBROUTINE statement (and not the TYPE declaration statement) as probably being easier to parse.

Discussing the implications of the constraints on intrinsic assignment, 10.2.1.2, their interpretation, 10.2.1.3, and the constraints on defined assignment, 10.2.1.4, and their interpretation, 10.2.1.5 on default array assignments.

Briefly discussing the implications of an argument with the DEFAULT attribute also having any one of the POINTER, VOLATILE, ASYNCHRONOUS, or TARGET attributes.

### DIFF
--- a/proposals/default_optional_arguments/revised_proposal.txt
+++ b/proposals/default_optional_arguments/revised_proposal.txt
@@ -1,0 +1,356 @@
+To: J3                                                     J3/XX-XXX
+From: M. Curcic & J. Vandenplas & Z. Jibben & W. Clodius 
+Subject: Default values for arguments with INTENT(IN), VALUE, or no
+         intent attributes
+Date: 2020-January-10
+Reference: 18-122r1, 18-136r1, 20-107
+
+
+1. Introduction
+
+This paper contains a proposal for Fortran 202y, to allow a programmer
+to specify a default value for a subset of optional dummy
+arguments. This would allow the programmer to then safely reference 
+such arguments in expressions regardless of whether the actual
+argument is present or not.
+
+
+2. Problem
+
+Fortran 2018 does not allow setting a default value for optional
+arguments. A default value is the value that the dummy argument
+would take if the corresponding actual argument is not present. If a
+dummy argument of INTENT(IN) or VALUE, or no intent is declared as
+optional, the user must:
+
+  * Explicitly test for the presence of the actual argument using the
+    intrinsic function present();
+  * Use a separate variable inside the procedure to assign the value
+    because the optional dummy argument that is not present must not
+    be referenced in expressions other than as actual argument to the
+    intrinsic function present().
+
+This example function illustrates the problem:
+
+    real function quadratic(x, a, b, c)  
+      ! returns a + b * x + c * x**2 if c is present 
+      ! and a + b * x otherwise
+      real, intent(in) :: x, a, b
+      real, intent(in), optional :: c
+      real :: c_tmp ! use another var. to reference the missing arg
+      c_tmp = 0 ! default value if c is not present
+      if (present(c)) c_tmp = c
+      quadratic = a + b * x + c_tmp * x**2
+    end function quadratic
+
+For any dummy argument with the optional attribute, the programmer
+must use the intrinsic function present() to check for the presence of
+the argument. Furthermore, for INTENT(IN) or VALUE or no intent
+arguments, if the optional dummy argument is meant to be used in
+multiple places in the procedure, the programmer is likely to use the
+above pattern, where a "temporary" variable is declared and used in
+place of the dummy argument, disconnecting the implementation from the
+user interface. Furthermore, this requires at least 3 lines of code
+(declaration of c_tmp, initialization of c_tmp, and testing for the
+presence of c) to handle the scenario of a missing optional argument.
+
+The situation for other intents is different. For INTENT(OUT) one
+might like to have a default value, but there is no need for a
+temporary variable to hold it. The example subroutine shows a typical
+usage:
+
+    subroutine do_stuff( ..., status )
+	...
+	integer, intent(out), optional :: status
+	...
+	if (present(status)) status = 0 ! the desired default
+	...
+	if (bad_stuff_happened) then
+	    if (present(status)) then
+	        status = 1 ! Flag that bad stuff happened
+		return
+	    else
+	        error stop 'Bad Stuff Happened in DO_STUFF'
+            end if
+	end if
+	...
+    end subroutine do_stuff
+
+This differs from the INTENT(IN), VALUE, and no intent case: first,
+because one assigns the default value with INTENT(OUT) when the actual
+argument is present, not when it is absent; and second, because the
+assignment of a default involves only one statement rather than the
+three statements for the helper variable of INTENT(IN), VALUE, and no
+intent. For INTENT(INOUT), one typically makes the argument optional
+to avoid the computational expense of calculating the out result, but
+having a default value makes it awkward to determine the presence of
+an optional argument, so the primary purpose of having it be optional
+is defeated.
+
+This proposal addresses the issue for INTENT(IN), VALUE, and no intent
+arguments that checking for the presence of the optional dummy
+argument and using a helper variable is cumbersome and error-prone. We
+will not address default values for INTENT(OUT) OR INTENT(INOUT)
+arguments. The primary benefit of this feature is the  reduction in
+source code needed to handle optional arguments. This benefit is
+greatest in scenarios where the optional argument is used in many
+places in the procedure, and a helper variable is used for its value
+instead. Reduction in needed source code would result in more readable
+and more correct programs.
+
+
+3. Prior art
+
+We are not aware of any Fortran compiler that has implemented default
+arguments. However it is a very common feature of other languages. Of
+interpreted languages widely used in the scientific community:
+Python, IDL, R, and Matlab all have default arguments. Of compiled
+languages C++ and Ada may be the best known with default arguments. Of
+these languages, Ada may be the most pertinent as it has INTENT
+arguments similar to Fortran's. Ada has only defined default arguments
+useful for INTENT(IN), and has not defined it for INTENT(OUT) or
+INTENT(INOUT).
+
+While the majority of default argument usages in the code we have
+seen use the equivalent of constant expressions for the default
+assignment, all of the above languages with default arguments allow
+some form of non-constant expressions for default arguments. For the
+interpreted languages it is up to the user to ensure that the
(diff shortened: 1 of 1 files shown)


certik:
Thanks for submitting this!

I like this. The only thing I am not sure is to specify the default expression in the subroutine statement. Indeed that is how other languages do that, but those languages also define the type there. In Fortran we define types separately from the subroutine line. So I think the default value should be initialized at the declaration also.

Note also that in your proposal the default keyword is redundant. If you remove it, you arrive at the initial proposal from Milan. See the discussion at the original issue.

wclodius2:
OTOH I suspect that overriding the default assignment will normally be done by “assignment" to the named argument in the procedure call. The similarity to the usage I find attractive. I also find it very awkward attempting to distinguish between the constant assignment that gives the save attribute to local variables, and the proposed default assignment to arguments. The definition of an entity declaration in the TYPE declaration statement will become even more complex. 

The DEFAULT keyword is redundant in the same sense that the OPTIONAL keyword is redundant. Fortran could have implicitly assigned the default attribute to any argument that is used with the PRESENT() function. I find the DEFAULT keyword very useful in distinguishing between arguments with the DEFAULT and OPTIONAL attributes. Redundancy isn’t always bad. See what happened to the SAVE attribute.

wclodius2:
After checking mail y Ada textbook, I found that one thing I wrote about Ada is wrong so I am going to correct the proposal.

wclodius2:
Removed the comment implying Ada was the exception in having a specification part.

klausler:
> Restricting optional arguments to arguments with the INTENT(IN), VALUE, or no intent attributes.

This would invalidate existing code, yes?

proposals/default_optional_arguments/revised_proposal.txt
```diff
@@ -0,0 +1,356 @@
...
+such arguments in expressions regardless of whether the actual
+argument is present or not.
+
+
+2. Problem
+
+Fortran 2018 does not allow setting a default value for optional
+arguments. A default value is the value that the dummy argument
+would take if the corresponding actual argument is not present. If a
+dummy argument of INTENT(IN) or VALUE, or no intent is declared as
```
jvdp1, line 23: Is `no intent` equivalent to `intent(inout)`?

proposals/default_optional_arguments/revised_proposal.txt
```diff
@@ -0,0 +1,356 @@
...
+intent. For INTENT(INOUT), one typically makes the argument optional
+to avoid the computational expense of calculating the out result, but
+having a default value makes it awkward to determine the presence of
+an optional argument, so the primary purpose of having it be optional
+is defeated.
+
+This proposal addresses the issue for INTENT(IN), VALUE, and no intent
+arguments that checking for the presence of the optional dummy
+argument and using a helper variable is cumbersome and error-prone. We
+will not address default values for INTENT(OUT) OR INTENT(INOUT)
```
jvdp1, line 93: I am not sure to understand why `intent(inout)` is not considered.If no argument is passed, then `intent(inout)` would behave as `intent(in)`?

proposals/default_optional_arguments/revised_proposal.txt
```diff
@@ -0,0 +1,356 @@
...
+3. Prior art
+
+We are not aware of any Fortran compiler that has implemented default
+arguments. However it is a very common feature of other languages. Of
+interpreted languages widely used in the scientific community:
+Python, IDL, R, and Matlab all have default arguments. Of compiled
+languages C++ and Ada may be the best known with default arguments. Of
+these languages, Ada may be the most pertinent as it has INTENT
+arguments similar to Fortran's. Ada has only defined default arguments
+useful for INTENT(IN), and has not defined it for INTENT(OUT) or
```
jvdp1, line 111: I don't know Ada. What does it implement for "no intent"?

milancurcic:
@klausler I'm pretty sure @wclodius2 meant "Restricting default arguments...". This proposal won't invalidate existing code. If it does, we will fix it.

jvdp1:
> This would invalidate existing code, yes?

After having read the document, you should indeed read "Restricting default arguments...".

wclodius2:
A thinko. The proposal talks about restricting **default** arguments to arguments with the INTENT(IN), VALUE, or no intent attributes, which doesn’t invalidate existing code.

wclodius2:
Unfortunately no intent is not equivalent to intent(inout), though I wouldn’t object strongly if the revised proposal dropped allowing default assignment for no intent. INTENT(INOUT) means that the input value is always intended to be used and an output value is always expected. No INTENT could behave as an INTENT(IN), INTENT(OUT), or INTENT(INOUT), and its behavior could depend dynamically on the other arguments.

jvdp1:
> Unfortunately no intent is not equivalent to intent(inout), though I wouldn’t object strongly if the revised proposal dropped allowing default assignment for no intent. INTENT(INOUT) means that the input value is always intended to be used and an output value is always expected. No INTENT could behave as an INTENT(IN), INTENT(OUT), or INTENT(INOUT), and its behavior could depend dynamically on the other arguments.

Thank you for the precisions. Therefore, because "no intent" could behabe as an `intent(out)`, wouldn't it make sense to not allowing default assignment for "no intent"? Otherwise, users could just drop "intent(out)" and add a `default` value for an argument actually behaving as `intent(out)`

wclodius2:
Like I said I have no strong feelings about no intent. The users can do a lot of things in almost any programming language that I would consider foolish. With no intent the user could also write the code in such a way that when he provides no argument so the default is used, the routine behaves like an INTENT(IN) and behaves like INTENT(INOUT) otherwise.. Whatever we decide we should discuss the implications for no intent, either requiring that an argument with the DEFAULT may not have the INTENT(INOUT) or INTENT(OUT) (implicitly allowing no intent), or require that arguments with the DEFAULT attribute must have  one of the attributes INTENT(IN) or VALUE (implicitly disallowing no intent).

proposals/default_optional_arguments/revised_proposal.txt
wclodius2: Ada doesn't have that option. For its parameters (the Fortran arguments) you always have to specify INTENT with either IN, OUT, or INOUT.

certik:
I don't like specifying the default value in the argument list as in `real function quadratic(x, a, b, c=0)`. I am ok with using `default` (and not mix this with `optional`).

One objection from the committee I believe is that the syntax `integer, intent(in), default :: c = 0` implies the `save` attribute. So perhaps this syntax should be used instead: `integer, intent(in), default(0) :: c`.

milancurcic:
> One objection from the committee I believe is that the syntax integer, intent(in), default :: c = 0 implies the save attribute.

I don't understand this objection because dummy arguments can't have a save attribute.

I am also okay with not mixing `default` and `optional`. I slightly prefer the `default :: c = 0` syntax, but `default(0) :: c` is fine too.

klausler:
You need to decide whether the default value of a dummy argument (or just that fact that it has one) is meant to be a characteristic of the procedure or not.  This matters in the implementation, because it determines whether an implementation has the freedom to implement the default value in the caller (as in C++) or in the called subprogram, or whether the language is forcing one or the other.

Why this matters: if the default value of a dummy argument is an expression involving other dummy arguments, host-associated variables, module or COMMON objects, &c., implementing the evaluation of the default value in the caller can be difficult or impossible.  But if default values are expressions to be evaluated if need be in the called subprogram, then the value need not be part of the characteristics or interface, and the implementation of the calling side is no different from an `OPTIONAL` argument; and one could argue that an explicit interface need not distinguish between `OPTIONAL` and having a default value for purposes of procedure pointer assignment and other interface compatibility checking.

certik:
Peter, when I discussed that at the last meeting (after the plenary), I think our idea was to implement the default value in the caller, which enables the compiler to compile the subroutine ahead of time without any if statements, and call it exactly the same way no matter if the user provides the default argument or not.

You have a good point about what if the default value is an expression involving other dummy arguments that it might be impossible. I don't know yet if it is impossible, and so a prior compiler implementation would be in order here. At least in gfortran it seems any of these expressions (such as a length of an array) is part of the function signature, so this would be also.

klausler:
> Peter, when I discussed that at the last meeting (after the plenary), I think our idea was to implement the default value in the caller, which enables the compiler to compile the subroutine ahead of time without any if statements, and call it exactly the same way no matter if the user provides the default argument or not.
> 
> You have a good point about what if the default value is an expression involving other dummy arguments that it might be impossible. I don't know yet if it is impossible, and so a prior compiler implementation would be in order here. At least in gfortran it seems any of these expressions (such as a length of an array) is part of the function signature, so this would be also.

```
subroutine host(defaultY)
  real, intent(in) :: defaultY
  call solver(f)
 contains
  real function f(x, y)
    real, intent(in) :: x, y
    default(defaultY) :: y
    ...
  end function
end subroutine
```

Seems like a compelling use case to me, but the default has to be implemented in the called subprogram to make it work.  So there's reasons why defaults should be implemented in callees, and the only argument for implementing them in callers (it might save a predictable branch) is weak and also has other obvious solutions.

certik:
I see. The nested subroutine `f` has access to the variables in the parent scope, such as `defaultY`, so this would work if the default is *not* part of the signature, using the current mechanism for nested subroutines. But if it is part of the signature of `f`, then one needs access to `defaultY` inside the function `solver`, which currently is not possible.

klausler:
> I see. The nested subroutine `f` has access to the variables in the parent scope, such as `defaultY`, so this would work if the default is _not_ part of the signature, using the current mechanism for nested subroutines. But if it is part of the signature of `f`, then one needs access to `defaultY` inside the function `solver`, which currently is not possible. Good point, thank you @klausler.

Here's an idea: extend the `VALUE` attribute to accept an expression, and allow it to be applied to an `OPTIONAL` dummy argument as its default value.   (You could also accept `VALUE` with an expression to specify a per-invocation initializer for a local variable, too, I suppose.)

PR #176, opened 2020-07-20
Add namespace modules proposal revision by @wclodius2

milancurcic:
Uploaded for @wclodius2. Related to issue #1.

### DIFF
--- a/proposals/namespace_modules/20-xxx.txt
+++ b/proposals/namespace_modules/20-xxx.txt
@@ -1,0 +1,176 @@
+To: J3                                                     J3/XX-XXX
+From: Ondrej Certik & William B. Clodius
+Subject: Namespace For Modules
+Date:
+#Reference: J3/19-246.txt, J3/20-108.txt
+
+Proposal for Fortran Standard: 202y (NOT 202x)
+
+
+1. Introduction
+
+The proposal is to allow the import of a module as a namespace to
+require accessing its members using the % operator. Example:
+
+    use, with :: utils
+    ...
+    call utils%savetxt(...)
+
+Where `utils` is the only name that is imported in the local
+namespace. `savetxt` is not accesable directly, only via `utils%`.
+
+This proposal originated at the J3 GitHub repository at [1].
+
+2. Motivation
+
+Several languages with the equivalent of Fortran's modules, e.g.,
+Python, Ada, and Haskell, have the ability to limit access to a
+module's members to a syntax equivalent to module_name %
+member_name. They allow this restriction first, to avoid name
+conflicts for members defined in different large modules, and second,
+to document locally the origin of a member as the origin may have
+implications for the detailed behavior or interpretation of a
+member. The users of these languages find the ability to make these
+restrictions useful.
+
+Of the three languages cited, the most widely used is Python so we
+will use it for our examples. Fortran module usage is equivalent to
+Python's:
+
+    Python                             Fortran
+
+    from A import foo                  use A, only: foo
+    from A import foo as Afoo          use A, only: Afoo => foo
+    from A import *                    use A
+
+Except:
+
+    Python                             Fortran
+
+    import A                           N/A
+    import A as B                      N/A
+
+This proposal proposes to fill in the missing functionality as
+follows:
+
+    Python                             Fortran
+
+    import A                           use, with :: A
+    import A as B                      use, with :: B => A
+
+or perhaps as
+
+
+    Python                             Fortran
+
+    import A                           with A
+    import A as B                      with B => A
+
+
+3. Use Cases
+
+3.1 Same function names in multiple modules
+
+In Python a very common idiom is:
+
+    import math
+    import numpy as np
+    import sympy as sym
+    ...
+    e1 = np.sin(np.pi)      # NumPy expression
+    e2 = math.sin(math.pi)  # Built-in Python math expression
+    e3 = sym.sin(sym.pi)    # SymPy expression
+
+In Fortran currently one has to do:
+
+    use math, only: math_sin => sin, math_pi => pi
+    use numpy, only: np_sin => sin, np_pi => pi
+    use sympy, only: sym_sin => sin, sym_pi => pi
+    ...
+    e1 = np_sin(np_pi)      ! NumPy expression
+    e2 = math_sin(math_pi)  ! Built-in Python math expression
+    e3 = sym_sin(sym_pi)    ! SymPy expression
+
+With this proposal one could also do:
+
+    use, with :: math
+    use, with :: np => numpy
+    use, with :: sym => sympy
+    ...
+    e1 = np%sin(np%pi)      ! NumPy expression
+    e2 = math%sin(math%pi)  ! Built-in Python math expression
+    e3 = sym%sin(sym%pi)    ! SymPy expression
+
+
+3.2 Need to import lots of functions from a module
+
+Existing code (https://github.com/certik/fortran-utils/blob/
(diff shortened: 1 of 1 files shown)


proposals/namespace_modules/20-xxx.txt
```diff
@@ -0,0 +1,176 @@
...
+    use strings, only: string_vector => vector
+
+    type(math_vector) :: avector
+    type(string_vector) :: bvector
+    ...
+
+vs.
+
+    use, with :: math, only: vector
+    use, with:: strings, only: vector
```
certik, line 146: I believe this should be just:
```suggestion
    use, with :: math
    use, with :: strings
```
the code below would still run. I don't think we should allow `user, with :: math, only: vector`, what would it do? Allow `math%vector`, but not `math%matrix`? Python doesn't allow that, I wouldn't allow that either here.

wclodius2:
FWIW I don’t have a significant  problem with disallowing it, but it requires more changes in the standard than simply allowing it,. In allowing the syntax you want to disallow you need only
1. Add the new keyword, with;
2. Add the module rename list; and
3. Disallow the module rename list if with is omitted; 

To disallow the syntax you want to disallow you need to add, probably as constraints:
4. Disallow the only keyword if with is present; and
5. Disallow the entity rename list if with is present.

It become a different negotiation i instead of use, with :: … you add a new with::... statement.  With a with statement you need:
1. Add the new keyword with
2. Add the with statement to the syntax
3. Have the module rename list in the statement

You are in effect adding a simpler statement rather than complicating an existing statement

certik:
I personally do not like allowing `use, with :: math, only: vector`. Also, if it is allowed, I think the proposal needs to explain more clearly what it would do. I personally would not allow it.

But I am happy to merge it if you just want to keep the proposal as is and discuss at the committee.

wclodius2:
Then put it on the back burner. There is no rush for F202y. My attitude is that if there is a lot of activity on an F202Y issue on the GitHub forum, then it probably is not ready for J3. If there is not a lot of activity then we need to ask why. If that is because everyone is happy with the proposal then get it to J3 as soon as possible so non-participants in Git Hub can comment and some work can be done between meetings, even if it is not F202X work. If it is because people are tired of the issue or busy with other things then it may not be ready. As you are both the originator of the first namespace proposals, and the organizer of the Git Hub activities you have a big say as to whether it is ready.

For an F202X issue if there is a lot of activity on the GitHub forum then I would put off submitting it to J3 until about two weeks before the meeting. It means that progress is being made at the local level, and you want to make as much progress as possible before involving all of J3, but you cannot put off involving J3 until after the meeting. The face to face involvement at meetings can resolve issues faster than email exchanges.

septcolor:
A very peripheral question, but for the J3 meeting, is it usual that participants are given only two weeks for reading proposals and making opinions about it? I don't think it is a sufficiently long period of time to prepare opinions. (CC: @sblione)

sblionel:
@septcolor The two-week limit is for new papers, though given that J3 meetings often generate dozens of new papers on the fly, it's not clear to me that this matters. Typically papers trickle in over the time between meetings so that two weeks in advance there are 20-30 papers to review (some of which have been there for months.)

We're at an odd stage in the development in that there are only seven work items for which work remains to be done. These all have topics in the j3-fortran.org discussion board (for J3 members only). So far, only a couple of minor papers have been submitted, none about the work items. I have been encouraging members to draft papers, but many are finding it difficult to devote time to this. If this in relation to 202Y proposals (as @wclodius2 suggests), then I suggest doing all discussion here as, at least for the October meeting, which is abbreviated, we may not have a chance to discuss future ideas.

wclodius2:
@sblionel so standards development is at an odd state. Enough work has been done on the current draft that there would normally be enough time available to start preliminary work identifying features for the following draft, but because people are having trouble adjusting to COVID-19 telecommuting it is not clear how much progress will be made on the current draft. As a result it is unlikely that any work will be done this session on F202Y, with the possible exception of generics/templates. Understandable, but disappointing.

wclodius2 (approved):

certik:
I've been thinking about this.

I don't know if I like `with` more than `namespace`. For sure `with` is easier to type. @wclodius2 can you motivate how `with` describes that it imports the `module` as a namespace?

proposals/namespace_modules/20-xxx.txt
wclodius2: FWIW I tried going to the file accept the change and nothing seemed to happen. I wonder if Milan has ownership.

certik:
My other objection is that I don't understand what `use, with :: math, only: vector` means. As I suggested, I would not allow it. (I understand your argument about the simplest possible change, but I feel that's secondary.)

wclodius2:
There are several possible syntaxes for a namespace requiring statement. I will use with for the examples, as it is shorter, but replace it with namespace if you like.

what we have been using for our examples is

use, with :: module-name

but that interacts oddly with module-nature

use, with, intrinsic:: iso_fortran_env
use, intrinsic, with :: iso_fortran_env

and you have to explicitly forbid rename-list and only-list

You could also put the with after the module-name
use [, module-nature  :: ] module-name, with

where with in effect replaces only making a third kind of use-stmt. This is easier to describe than the form we have been using for our examples.

Finally you could have a namespace statement

with [, module-nature ::] module-name

It is what Ada uses, it is more succinct. It is also easier to  describe than the form we have been using for our examples.

wclodius2:
Using a module as a namespace you use the module entities always with the module-name.

certik:
> Using a module as a namespace you use the module entities always with the module-name.

Yes. I am just trying to understand how the term `with` describes this. You didn't use `with` in the above sentence, but you used the term `namespace`.

wclodius2:
I am having trouble understanding your point. I did use with in the sentence. I could have just as easily left namespace out.

Using a module with the with attribute requires that a module entity always be named with the module-name prepended.

certik:
My apologies, I amended my comment at https://github.com/j3-fortran/fortran_proposals/pull/176#issuecomment-696479777, yes you did use `with`. 

Will you be able to join our Fortran call this Friday: https://fortran-lang.discourse.group/t/fortran-monthly-call-september-2020/298/8

Let's discuss this there.

Let me try to reformulate my question: I believe that language "keywords" should be natural and intuitive to describe what the feature is doing. So I want to have a motivation. Here is my best effort to motivate the two options:

* `use, namespace :: math` The `namespace` keyword comes from: The `math` module will be imported as a **namespace**.
* `use, with :: math` The `with` keyword comes from: The `math` module will be imported as a namespace where you use the module entities always **with** the module-name.

But I also have another alternative motivation:

* * `use, with :: math` The `with` keyword comes from: The code that contains this line will be used **with** the `math` module.

Which of these (if any) make sense to you?

wclodius2:
I will try to attend the the Fortran call on Friday. I currently have no conflicts with that time, but that of course can change.

> use, with :: math The with keyword comes from: The use of the module entities is always with the module-name.

Leonard-Reuter:
What is the current state of this proposal? I think it is a desperately needed feature. 
Regarding the discussion over here: I'd also prefer `namespace` over `with`. Anybody new to the language will easily understand what the former does, while it requires some guess work for the latter.

sblionel:
It's on the list for consideration for Fortran 202Y. The Data subgroup's initial comments are:

3. Namespace for modules
    The proposals here are multitudinous; much of the functionality is
    already provided by existing syntax, the main thing that looks useful
    would be a syntax for "remote access" to module entities, e.g.
    "modulename % entityname". Not sure % is the best thing for this.
Medium desirability, Limited scope, Fairly easy.

PR #177, opened 2020-08-04
Assert

wclodius2:
A paper on design by contract that, among other things, proposes the addition of an ASSERT statement
to Fortran 202y.

### DIFF
--- a/proposals/assert/design_by_contract.txt
+++ b/proposals/assert/design_by_contract.txt
@@ -196,7 +196,7 @@ Constraint: requires-expr shall be a restricted expression.
 If thought of as run time expressions then the appropriate syntax may
 be
 
-  requires-stmt is EEQUIRES( requires-expr-list )
+  requires-stmt is REQUIRES( requires-expr-list )
   ensures-stmt is ENSURES( ensures-expr-list )
 
 In all of the syntactical forms each item in the effective
--- a/proposals/assert/design_by_contract.txt
+++ b/proposals/assert/design_by_contract.txt
@@ -1,0 +1,243 @@
+To: J3                                                     J3/XX-XXX
+From: William B. Clodius
+Subject: Facilitate design by contract
+Date: August 1, 2020
+#Reference: 04-142.txt, 04-143.txt, 04-144.txt, 04-212.txt,
+04-219.txt, 04-376.txt, 04-414.txt
+
+Proposal for Fortran Standard: 202y (NOT 202x)
+
+1. Introduction
+
+This paper contains a proposal to allow programmers to use "Design by
+Contract". Inspired by languages such as C++, Java, and Ada, Fortran
+is considering incorporating exception handling into the language. In
+contrast, C++'s standard committees have decided that its users have
+been overusing exception handling, and has been seeking to incorporate
+other, simpler, error handing systems.[1] One approach, that has been
+incorporated into C++ 2020, is design by contract.[2,3] I believe
+Fortran would benefit by incorporating design by contract into the
+language, at a cost much less than would be incurred by incorporating
+exception handling.
+
+Design by contract (DbC) is intended to address programmer errors and
+only those errors. The idea is to specify: first the logical
+preconditions the inputs to a procedure are required to satisfy for
+its correct operation; second, the conditions the code is asserted to
+satisfy at various stages of its logic if the code logic is correct,
+and finally, the post-conditions the code ensures it will satisfy if
+the pre-conditions are met. To implement these checks, C++ 2020 has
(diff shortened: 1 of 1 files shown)


proposals/assert/design_by_contract.txt
```diff
@@ -0,0 +1,247 @@
...
+
+  requires-expr is logical-expr
+  ensures-expr is logical-expr
+
+Constraint: requires-expr shall be a restricted expression. 
+
+If thought of as run time expressions then the appropriate syntax may
+be
+
+  requires-stmt is EEQUIRES( requires-expr-list )
```
milancurcic, line 199:
```suggestion
  requires-stmt is REQUIRES( requires-expr-list )
```

proposals/assert/design_by_contract.txt
```diff
@@ -0,0 +1,247 @@
...
+
+where
+
+  assertion is logical-expr
+
+REQUIRE and ENSURES can be thought of as either sub-program attributes
+or executable statements. If treated as sub-program attributes then
+the syntax could be something like
+
+  requires-stmt is EEQUIRES :: requires-expr-list
```
milancurcic, line 188:
```suggestion
  requires-stmt is REQUIRES :: requires-expr-list
```

milancurcic: @wclodius2 If you like the edits, you can accept them by clicking on "Commit suggestion".

milancurcic:
Thank you @wclodius2, I read the proposal. Do `requires` and `ensures` do anything different from `asserts` other than being required to be in the prelude and conclusion of a procedure, respectively? Why not use just `asserts`?

I have a slight preference for imperative mood for statements like this (`require` instead of `requires`, `assert` instead of `asserts`), which is consistent with other Fortran statements (`allocate`, `deallocate`, `nullify`, `open`, `close`, `read`, `write`, `inquire`, `change team`, `form team`, etc.)

nncarlson:
I would find it useful to have an additional statement `INSIST` that behaves identically to `ASSERT` except that it is active in both development and deployment compilation modes; i.e., it is never a null operation.

wclodius2:
The change is good.

wclodius2:
Having them be different from assert calls them out to users of the procedure, and to automatic processing codes such as FORD. They naturally suggest documentation such as ASIN(X)  requires X>= -1, and X<=1, and ensures the result is >= -pi/2 and <= pi/2. The requires and ensures are aspects of the public interface of a procedure, and are important to users of the procedure, while assert is important only to the developer of the procedure.

wclodius2:
FWIW I slightly prefer the active voice.

wclodius2:
I don't seem to have write access to the repository, so I have to ask others to make changes for my approval. As it is always an executable construct ASSERT should always be in the active voice. Can someone change all appearances of ASSERTS to ASSERT? When discussed as an attribute statement I believe I should use REQUIRES and ENSURES, while when discussed as an executable statement I believe I should use REQUIRE and ENSURE. Can someone make that selected change?

nncarlson:
@wclodius2  I'm not sure what errors you're seeing, but you should be able to make changes to your pull request.  Looks like Milan committed a couple changes to it. So perhaps you need to do a `git pull` on your assert branch to pull those changes into your working copy before you can commit your changes and push them back to the assert branch on your fork of the repo.

wclodius2:
It has been a month and none of my three proposals (assertions/design by contract; optional dummy arguments, and module namespaces) have been merged into the repository. Did they "fall off the radar"? Did they need more work? Are they inappropriate for meeting 222, as they are explicitly F202y oriented and so low priority? Are there other reasons I am not thinking of?

certik:
@wclodius2 in the past we usually merged the pull requests before the meeting once we were happy with the wording and it was the version we wanted to present. I am happy to merge it sooner if you would prefer.

In the last meeting we presented such proposals at the plenary to get feedback from the whole committee. These proposals go into 202y, so the committee does not vote on them. The idea is to get them ready, so that when the committee starts considering proposals for 202y, they are ready.

In order to increase chances of the proposal getting accepted, it's good to involve as much of the wider community as possible. If there is wide support for some feature, it has a much higher chance of getting in.

wclodius2:
In the past the nominal due date for proposals was two weeks before the meeting, with most proposals arriving a few days before the schedule. This means most proposals were read in a rush in the two week period before the meeting. I think where possible F202X proposals should be submitted a month in advance so there is a likelihood of feedback before the meeting. As these are F202Y proposals they don’t need as prompt a feedback, but still I think they should get in before the two week deadline so there is a possibility of pre meeting commentary.

certik:
Ok, let's merge it.

Issue #156, opened 2020-02-25
Protected Components


certik:
As part of #155, the committee is looking at protected components as part of

US 19 Protected Components

https://j3-fortran.org/doc/year/19/19-135r1.txt
https://j3-fortran.org/doc/year/19/19-161.txt
https://j3-fortran.org/doc/year/19/19-214r1.txt

certik:
@zjibben can fill in the details.

It looks like the latest version so far has a requirement that a protected component (meaning one cannot write into it from outside the class) will require that one cannot assign the class variable to another (`a = b`), or allocate it or deallocate it, unless one writes lots of boilerplate code (custom initializer? and finalizer and custom `=` operator and the user must import it).

I am personally against that, I don't think that's a good approach. If using the equivalence of `=` being "copy constructor in C++", allocate being "constructor" and "deallocate" being "destructor", those in C++ can be implicitly generated by the compiler, or written explicitly by the user, in either case they are methods of the object itself, and thus if C++ had protected components, it seems that one could call `=` and create and destruct the object. I would do the same in Fortran.

klausler:
C++ does have `const` data members that can be assigned only during construction.

zjibben:
As @certik said, there is something in the works related to #16, but the existing version has many more restrictions than what we expected. In particular, if you have a type with a protected component (or some subobject which ultimately has a protected component), there are many things you wouldn't be able to do outside the module where it is defined. The restrictions which raised concern were, outside the module where the type containing a protected component defined, you could not allocate or deallocate a variable of the type, nor use intrinsic assignment to the object as a whole if it contains a protected data member (`a = b` is disallowed if `b%c` is protected). It also restricted type extension, so one could never have an extension type with a protected component if the parent didn't also have a protected component (this restriction would prevent a loophole to get around the no-intrinsic-assignment rule above).

Some committee members were concerned this would make protected components too unwieldy to use in practice. They would force a lot of boiler plate to use defined assignment or type extension or allocatables if all a programmer wanted was automatic and free getter methods. On the other hand, the counter-argument is that without these restrictions you aren't really protecting the data all that much. Anyone outside the module could blow it away with a reallocation or intrinsic assignment to the variable which ultimately contains protected data.

We agreed the restriction on type extension was too far, so that is going away. The rest is up for debate. At the moment things are leaning toward reworking the requirements entirely, that the original version may be too heavy a hammer, but I think nothing is really settled one way or another.

[...]

PR #182, opened 2020-09-21
Add revised protected components & types proposal

zjibben:
Here is a revised version of the protected components proposal (#156). The next committee meeting is 3 weeks away (Oct 12) so I wanted to make sure we had time to discuss beforehand.

To recap, last time it became clear there were competing interests. I and others here and at LANL wanted an access specifier roughly in between `private` and `public`. Other committee members want something far more restrictive, something really unlike an access specifier and more analogous to the existing `protected` attribute for module data. This would act to really protect the data from virtually any modifications outside the module where the type was defined, and was far too restrictive to be useful for those of us who wanted an access specifier.

The best way forward seemed to be to tease apart these separate needs and provide tools to manage both. That way we all get what we want. This paper has specifications & syntax for protected components and protected types based on our withdrawn 20-121. If we're happy with it, I can either submit it as a paper or as a work item for the DATA subgroup.

### DIFF
--- a/proposals/protected-components.txt
+++ b/proposals/protected-components.txt
@@ -1,0 +1,245 @@
+To: J3                                                     J3/XX-XXX
+From: Zach Jibben
+Subject: Protected components & types: specifications, & syntax
+Date: 2020-October-1
+Reference: 20-121 20-106 19-214r1 19-161 19-135r1 18-265
+
+
+1. Introduction
+===============
+
+This paper contains the formal specifications & syntax for protected
+components of a type, and protected types. This supersedes the
+specifications outlined by previous papers, as subgroup discussion revealed
+a more flexible feature set was needed to meet competing user requirements.
+
+Previous protected-components specifications fell into one of two camps.
+Papers 18-265 and 20-106 envisioned components inaccessible for *direct*
+modification outside the module in which the parent type was defined, but
+did allow a type containing a protected potential subobject to appear in a
+variable-definition context. These papers propose an access specifier
+roughly in-between PUBLIC and PRIVATE, by effectively prohibiting the name
+of a protected component from appearing in a variable-definition context,
+not protecting the variable itself. Other papers, 19-135r1, 19-161,
+19-214r1, and 20-121 offered stronger protection, protecting variables
+themselves from being modified by virtually any means outside the module
+where that component was defined.
+
+This paper aims to satisfy both parties by teasing apart the competing
+goals into two separate features: protected components, and protected
+types. Protected components provide an access specification allowing code
+outside the module defining the type to read, but not *directly* modify
+components. These components will not impose any restrictions on the type
+containing them, beyond anything that might be imposed by the analogous
+PRIVATE or PUBLIC access specifiers. Protected types may be used to protect
+the data from appearing in variable-definition contexts.
+
+The use cases for both features have been presented in previous papers
+(protected components 18-265, 20-106, 19-214r1; protected types 19-135r1).
+
+Although this is a formal syntax paper, the syntax will be defined by prose
+and by example, not by BNF, to aid comprehension.
+
+
+2. Specifications
+=================
+
+To help keep a record of how specs have changed and to aid discussion,
+specification labels are preserved from papers 20-121 and 19-214r1.
+
+
+2.1 Protected Components
+
+B. The name of a protected component shall not appear in a
+   variable-definition context, except within the module wherein its type
+   is defined.
+
+G. A protected component or a subobject of a protected component can only
+   be argument associated with INTENT(IN) dummy, even if the referenced
+   procedure is in the module in which the type is defined.
+
+H. A protected component or subobject of a protected component cannot be
+   the target in a pointer assignment outside the module, as that would
+   lose the protection.
+   - Here I think it would be valuable to somehow expose an immutable
+     reference, which could tie into the const-pointer proposals.
+
+I. A protected component or subobject of a protected component cannot be
+   explicitly allocated or deallocated except within the module in which
+   the type is defined.
+   - Otherwise it's too easy to bypass the protection.
+   - Allocating/deallocating a type containing a protected component is
+     still permitted, provided it's not subject to some other rule.
+   - Automatic deallocation on scope exit will occur as it would for any
+     unprotected component.
+
+J1. A dummy variable of a type with a protected component can be INTENT(IN)
+    or INTENT(INOUT) or INTENT(OUT) in any procedure anywhere, unless it's
+    subject to some other rule.
+
+L1. No intrinsic assignment to a protected component or subobject thereof,
+    outside the module in which the protected component is defined. (cf.
+    19-135r1)
+    - This does not prohibit intrinsic assignment to a type containing a
+      protected component. It is only meant to prevent the protected
+      component from being modified by name.
+
+M1. A function defined outside the module may have a result variable of a
+    type with a protected component.
+
+N. Structure constructor outside the module in which the type is allowed if
+   and only if no value is supplied for any protected component (otherwise
+   this would subvert the module's control over what values are acceptable
+   in the protected component).
+
+P1. Protected components are inherited through type extension.
+    - In this regard, PROTECTED is like PUBLIC, not PRIVATE.
+
+R. Type-wide default protected status, like PRIVATE, is provided. Default
+   is PUBLIC, unchanged from the current standard. PROTECTED statement
+   changes the default. Attributes in a component definition stmt confirm
+   or override the default.
+
+U. SEQUENCE and BIND(C) types shall not have protected components.
+   - This includes any level of component selection, because non-SEQUENCE
+     types are not allowed in SEQUENCE, and similarly BIND(C)).
+
+AA. A component may be PROTECTED or PRIVATE or PUBLIC, but not a
+    combination of more than one of these attributes.
+
+
+2.2 Protected Types
+
+A. A variable whose declared type is protected shall not appear in a
+   variable-definition context, except within the module wherein its type
+   is defined.
+
+C. A local variable whose declared type is protected is allowed outside the
+   module in which the type is defined.
+   - If some action is needed when the variable goes out of scope (and is
+     thus destroyed unless it has the SAVE attribute) its type should have
+     a final procedure.
+
+D. A local pointer of a protected type is allowed outside the module in
+   which the type is defined.
+   - Otherwise use case 2.1 in 19-135r1 is not satisfied.
+
+E. Deallocating an object with a protected declared type outside the module
+   in which the type is defined is prohibited.
+   - Otherwise for pointer, use case 2.1 in 19-135r1 is not satisfied.
+   - Allowing it for remote/dummy allocatable would prohibit an allocatable
+     variable in the module where the type is defined to be PUBLIC.
+     PROTECTED attribute for the object would prevent modifying
+     nonprotected components.
+   - Allowing for local allocatable breaks rule (A) about not appearing
+     in a variable definition context.
+
+F. Allocating an object with a protected declared type outside the module
+   in which the type is defined is prohibited.
+   - Pointer would almost certainly leak memory.
+   - Allowing for remote/dummy allocatable would have same problems as (E).
+   - Allowing for local allocatable would have same problems as (E).
+
+J2. A dummy variable of a protected type can be INTENT(IN) or INTENT(INOUT)
+    in any procedure anywhere. A dummy variable of such a type shall not
+    have INTENT(OUT) or unspecified intent except within the module in
+    which the type is defined.
+
+K. Can modify (or allocate/deallocate) the *components* of a protected type
+   anywhere, provided the component is not subject to some other rule
+   (e.g., a protected component, or the component is itself of protected
+   type).
+
+L2. Polymorphic allocatable assignment of a parent type without a protected
+    attribute is permitted even when the dynamic type has a protected
+    attribute.
+   - This presents a loophole which may allow a programmer to write to a
+     protected component outside the module in which it is defined.
+     However, subgroup did not like the alternative of disallowing
+     protected components in extensions of parent types without any
+     protected component.
+
+L3. No intrinsic assignment to an object whose declared type is protected,
+    or that has an ultimate component of protected type, outside the module
+    in which the protected component is defined. (cf. 19-135r1)
+
+M2. A function defined outside the module may have a result variable of a
+    protected type.
+    - This is same as ordinary local variables (case A).
+    - The function result shall not be a pointer.
+    - The function result will (hopefully) be finalised after its use.
+
+N2. Structure constructor outside the module in which the type is allowed.
+
+O. Another type may have a component of protected type, and thereby
+   inherits the protected type attribute.
+   - The protection rules apply to types with "protected potential
+     subobject components" not just types with immediate protected
+     components.
+
+P2. Extension of a protected type is permitted.
+
+Q. An extension type may have a protected attribute even if the parent type
+   does not have a protected attribute. Requiring the parent type to have
+   be protected is too restricted for many uses.
+
+S. An object with a protected declared type is prohibited to be the target
+   of an unlimited polymorphic pointer.
+   - This rule applies even inside the module, even if the polymorphic
+     pointer is private, because its target might thereafter become
+     associated with a public polymorphic pointer.
+
+T. If an actual argument with a protected declared type is associated with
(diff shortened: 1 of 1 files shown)


zjibben:
@aradi

proposals/protected-components.txt
```diff
@@ -0,0 +1,245 @@
...
+      protected component. It is only meant to prevent the protected
+      component from being modified by name.
+
+M1. A function defined outside the module may have a result variable of a
+    type with a protected component.
+
+N. Structure constructor outside the module in which the type is allowed if
+   and only if no value is supplied for any protected component (otherwise
+   this would subvert the module's control over what values are acceptable
+   in the protected component).
```
wclodius2, line 93: I can't parse this sentence. Did you mean "A structure constructor outside the module, in which the type is defined, is allowed if and only if no value is supplied for any ...

proposals/protected-components.txt
```diff
@@ -0,0 +1,245 @@
...
+
+2.1 Protected Components
+
+B. The name of a protected component shall not appear in a
+   variable-definition context, except within the module wherein its type
+   is defined.
+
+G. A protected component or a subobject of a protected component can only
+   be argument associated with INTENT(IN) dummy, even if the referenced
+   procedure is in the module in which the type is defined.
```
wclodius2, line 59: with an INTENT(IN) dummy

everythingfunctional, line 59: I believe this is implied by B, but nothing wrong with being explicit.

proposals/protected-components.txt
```diff
@@ -0,0 +1,245 @@
...
+
+H. A protected component or subobject of a protected component cannot be
+   the target in a pointer assignment outside the module, as that would
+   lose the protection.
+   - Here I think it would be valuable to somehow expose an immutable
+     reference, which could tie into the const-pointer proposals.
```
everythingfunctional, line 65: Good eye, forward thinking. I had a similar thought, but didn't see a reason to bring it up and complicate matters. Worth keeping in mind though.

proposals/protected-components.txt
```diff
@@ -0,0 +1,245 @@
...
+
+    type, protected :: foo
+      real, public :: a
+      real, private :: b
+      real, protected :: c
+    end type foo
+
+The PROTECTED keyword meaning different things in these contexts may be
+confusing. It's worth considering different keywords (e.g., readonly
+components).
```
wclodius2, line 238: or visible components

proposals/protected-components.txt
everythingfunctional, line 93: I think we actually don't want this. The motivation behind the feature is more about keeping objects internally consistent, and a structure constructor would potentially (probably?) produce an internally inconsistent object.

proposals/protected-components.txt
```diff
@@ -0,0 +1,245 @@
...
+    or that has an ultimate component of protected type, outside the module
+    in which the protected component is defined. (cf. 19-135r1)
+
+M2. A function defined outside the module may have a result variable of a
+    protected type.
+    - This is same as ordinary local variables (case A).
+    - The function result shall not be a pointer.
+    - The function result will (hopefully) be finalised after its use.
+
+N2. Structure constructor outside the module in which the type is allowed.
```
everythingfunctional, line 172: Should this just be a copy of N?

proposals/protected-components.txt
zjibben: That's a good catch! Somehow this wording has survived 2 papers across 4 different authors, and nobody caught it. I'll fix it now.

proposals/protected-components.txt
zjibben: Thanks!

proposals/protected-components.txt
zjibben: Thanks. We'll see what comes of this, but given there's already work on the other feature it would be cool to get both of these working together.

proposals/protected-components.txt
zjibben: (I think) structure constructors are not modifiable. Of course a type-bound procedure called `init` or something like that can do anything, but this would be for the language-defined structure constructor. e.g.:

```
type :: foo
  integer :: a, b, c
  integer, private :: d
  integer, protected :: e
end type foo

type(foo) :: x

x = foo(a=1, b=2, c=3)
```

In this case, we can't specify values for `d` or `e` via the structure constructor. I'm sure some discussion will come up on this point, though, because a `private` variable `d` cannot be specified here because its name isn't accessible in the scoping unit. The name `e` is accessible, so some extra rules will need to be added to the standard to handle this.

certik (approved):
The paper looks good. 

My major suggestion for improvement would be to add test/example cases of how the feature is used. The only examples are how it is declared, but it would be nice to add examples showing:

* how it can be used
* what usage is allowed
* what usage is *not* allowed
* how exactly it differs from the previous proposals

proposals/protected-components.txt
zjibben: Now that I'm thinking about it, we might actually want the opposite: "A structure constructor outside the module in which the type is defined is *not* allowed".

The original version was concerned with assigning to a protected component, so prohibited intrinsic assignment but basically allowed an exception for structure constructors which did not name those protected components (though this could effectively overwrite existing data in a protected component, if I'm not mistaken). Now that we've separated out the protected components, I'm thinking we don't need the exception and would just forbid structure constructors on protected types (when outside the module where the type is defined).

I think I'll change it to that for now, but I'll defer to the interested parties in the committee for behavior of type protection.

proposals/protected-components.txt
zjibben: Thanks, I'll add that as another option. I think making clear that we're very open to other/more appropriate keywords will make it stronger.

proposals/protected-components.txt
everythingfunctional: Now I see it. `a`, `b`, and `c` aren't being protected anyways, so getting rid of the structure constructor doesn't buy you any extra protection. Got it.

aradi (commented):
Thx. for setting this up! This is all fine, see a few minor comments. My only concern is, that we are now putting two different categories of `protected`-nes into one proposal:
- The `protected` component, which is probably not too difficult to implement in compilers and seem to have not too many implications/side effects.
- The `protected` type, which seems to be a monster with many-many implications and side-effects.
I fear, that it would put the first one on hold due to numerous discussions on and problems with the second one.

Also, I'd recommend to use a different name for the two features as to me, their puprose is completely different. To me `protected` just means 'read-only', in the sense that you need routines of the defining modules to manipulate that value. Note, that also in case of `protected` module variables, the same meaning still holds. The only reason, why such variables are also behaving kind of 'persistent', is that modules do not go out of scope. If they would, `protected` module variables would "vanish" exactly the same way as all other (private and public) module variables, so they were not persistent. So it is not they `protected`-nes which makes they persistent... Therefore, I'd suggest to keep the term `protected` for expressing read-only behaviour (of the derived type components), and use something else for defining persistent/undestroyable types (e.g. `persistent`).

proposals/protected-components.txt
```diff
@@ -0,0 +1,245 @@
...
+J1. A dummy variable of a type with a protected component can be INTENT(IN)
+    or INTENT(INOUT) or INTENT(OUT) in any procedure anywhere, unless it's
+    subject to some other rule.
+
+L1. No intrinsic assignment to a protected component or subobject thereof,
+    outside the module in which the protected component is defined. (cf.
+    19-135r1)
+    - This does not prohibit intrinsic assignment to a type containing a
+      protected component. It is only meant to prevent the protected
+      component from being modified by name.
```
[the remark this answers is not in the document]
zjibben: I believe it is covered by B. The original paper I built off had both (19-214r1), and I can't say I know why. I'm taking a "change as little as possible" approach compared to what they had.

proposals/protected-components.txt
```diff
@@ -0,0 +1,245 @@
...
+   type).
+
+L2. Polymorphic allocatable assignment of a parent type without a protected
+    attribute is permitted even when the dynamic type has a protected
+    attribute.
+   - This presents a loophole which may allow a programmer to write to a
+     protected component outside the module in which it is defined.
+     However, subgroup did not like the alternative of disallowing
+     protected components in extensions of parent types without any
+     protected component.
```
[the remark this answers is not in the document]
zjibben: That's a good question, I'll clarify in here. This was something concluded by Malcolm & Van, and I believe the intention is that the protected component would become deallocated & finalized. This is basically a loophole to avoid the other restrictions on deallocation & finalization, but the alternative they had in mind (prohibiting protected components in type extensions) was too strong.

But I'm noticing some holdovers from the old protected component terminology, which here is a protected type. That needs to be fixed.

zjibben:
Thanks @aradi!

> The protected type, which seems to be a monster with many-many implications and side-effects.
I fear, that it would put the first one on hold due to numerous discussions on and problems with the second one.

This is definitely a valid concern. For our part, I wouldn't be proposing that monster feature at all (I don't see it as extremely useful personally). But when presented with protected components, the committee really wanted something extremely strong, and had lots of work towards that already. These specs derive from the work on those very restrictive protected components. It looks like if we want our more lightweight protected components, we need something strict to go with it. But we'll see as discussions develop, maybe seeing both alongside each other the committee will be willing to drop the heavy one. Hopefully the committee doesn't decide they don't see the value in our lighter protected components, deciding to go ahead with heavy-only, or decide the whole combination is more than they wanted and kill the whole concept.

> I'd suggest to keep the term protected for expressing read-only behaviour (of the derived type components), and use something else for defining persistent/undestroyable types (e.g. persistent).

I like this, and have struggled to come up with an alternative name for "protected type". `persistent` is a good option, as it really just disables intrinsic assignment, allocation, deallocation, and automatic finalization. I'll include that along with the other name alternatives I now have at the end. I think naming will be a big point of discussion.

zjibben:
@certik thanks for the suggestion. I added a section on use-cases, which naturally was very detailed on what we want to see in protected components, but lighter for protected types. I tried to faithfully relay the motivation described to me by Van, and I think it ended up being similar to described by Malcolm in 19-135r1.

zjibben:
Tagging a few others who might be interested: @FortranFan @milancurcic @gklimowicz @tclune @vansnyder

I think we got this into pretty good shape during our discussion in September, but I've added a bit more clarifying discussion. I'm planning to show this to the committee for our meeting next week. I think this approach offers enough flexibility to satisfy everyone, but we'll see how it goes.

vansnyder:
On Tue, 2021-06-15 at 15:20 -0700, Zach Jibben wrote:
> Tagging a few others who might be interested: @FortranFan
> @milancurcic @gklimowicz @tclune @vansnyder
> 
> I think we got this into pretty good shape during our discussion in
> September, but I've added a bit more clarifying discussion. I'm
> planning to show this to the committee for our meeting next week. I
> think this approach offers enough flexibility to satisfy everyone,
> but we'll see how it goes.
> 
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or unsubscribe.

Zach mentions that "persistent" types disable intrinsic assignment,
allocation, deallocation, and automatic finalization.

I assume this disabling does not apply in the scoping unit where the
type is defined.

In Ada '83, the "limited" attribute for a type did this. I don't
remember in which level of requirements this was first introduced. It
was definitely in "Steelman." I first proposed it in 97-114.

zjibben:
Hi Van, thanks for looking at this!

> I assume this disabling does not apply in the scoping unit where the type is defined.

This is correct, specification L3 allows for intrinsic assignment inside the module which defines the protected type.

FortranFan:
@zjibben wrote 15 June 2021, 6:19 PM EDT:

> I've added a bit more clarifying discussion. I'm planning to show this to the committee for our meeting next week. I think this approach offers enough flexibility to satisfy everyone, but we'll see how it goes.


Great paper, Zach.  I read your revised version closely just now and you've covered everything that I can think of, especially with formal specifications toward the **protected components** feature that is of particular importance in many codes in my experience.

My only change will be on line 113: 'inadvertently'.

I feel strongly this will be an excellent addition to Fortran 202X.  The discussion at the plenary on this paper will be of great interest, hope no technical surprises arise then.

proposals/protected-components.txt
```diff
@@ -0,0 +1,363 @@
...
+Although this is a formal syntax paper, the syntax will be defined by prose
+and by example, not by BNF, to aid comprehension.
+
+
+2. Use Cases
+============
+
+The use cases for both features have been presented in previous papers
+(protected components 18-265, 20-106, 19-214r1; protected types 19-135r1).
+Here is presented two possible use-cases. For more, refer to the above
```
milancurcic, line 46:
```suggestion
Here are presented two possible use-cases. For more, refer to the above
```

proposals/protected-components.txt
```diff
@@ -0,0 +1,363 @@
...
+      call new_mesh(m) ! passing object as intent(out) or intent(inout)
+                       ! allowed
+
+      m%nbr(1) = huge(1) ! FAIL: modifying protected component not allowed
+      call change(m%nbr(1)) ! FAIL: intent(out) and intent(inout) not
+                            ! allowed on protected components
+      m = mesh(nbr = [huge(1)]) ! FAIL: setting protected component
+                                ! via structure constructor not allowed
+      allocate(m%nbr(1)) ! FAIL
+      deallocate(m%nbr(1)) ! FAIL
```
jvdp1, line 94: Should it be:
```suggestion
      deallocate(m%nbr(:)) ! FAIL
```

proposals/protected-components.txt
```diff
@@ -0,0 +1,363 @@
...
+   - This describes the same user-facing behavior for PROTECTED and PRIVATE
+     components. However, the logic in the standard must be different.
+     C7102 and 7.5.4.8p2 prevent a structure constructor from specifying
+     a private component by restricting access to the name. For a PROTECTED
+     component, the name is accessible, yet we still prohibit setting the
+     value of such a component outside the module where the parent type is
+     defined.
+
+P1. Protected components are inherited through type extension.
+    - In this regard, PROTECTED is like PUBLIC, not PRIVATE.
```
jvdp1, line 176: In all other rules, PROTECTED behaves as PRIVATE, except for its readable behavior. Should it not be the case for this rule too?

jvdp1 (commented):
Thank you @zjibben. This feature would be a useful  addition. Good luck!

proposals/protected-components.txt
```diff
@@ -0,0 +1,363 @@
...
+above.
+
+
+5.3 Is a public component of a protected type any different from a
+protected component of a protected type?
+
+A: I believe the answer here is yes. Specification K allows modification of
+the components of a protected type; i.e. the protection of a type really
+only restricts what can be done to the type as a whole. This is up for
+discussion, though, and could be changed.
```
milancurcic, line 361: This seems reasonable to me.

proposals/protected-components.txt
```diff
@@ -0,0 +1,363 @@
...
+
+U. SEQUENCE and BIND(C) types shall not have protected components.
+   - This includes any level of component selection, because non-SEQUENCE
+     types are not allowed in SEQUENCE, and similarly BIND(C)).
+
+AA. A component may be PROTECTED or PRIVATE or PUBLIC, but not a
+    combination of more than one of these attributes.
+
+
+3.2 Protected Types
```
milancurcic, line 191: Can a protected type be extended? If yes, would the extending type also be protected? It seems to me that it would have to be, otherwise it would be easy to "unprotect" a protected type. This should be spelled out in the spec.

milancurcic (approved):
Great proposal and useful potential addition, thank you for pushing it.

proposals/protected-components.txt
milancurcic: Okay, I see P2.

proposals/protected-components.txt
zjibben: Good catch!

proposals/protected-components.txt
zjibben: My reasoning here is PROTECTED components are part of an API, something a client can see and read. PRIVATE components are completely invisible outside the type, hence it's safe not to inherit them. But if a client has a `class(parent)` with a protected component, the dynamic type `type(child)` ought to share the parent's API.

proposals/protected-components.txt
zjibben: Thanks Milan!

zjibben:
Thank you everyone! The paper has been submitted: https://j3-fortran.org/doc/year/21/21-148.txt

Issue #183, opened 2020-10-05
Conditional Expressions


certik:
Relevant papers:

* https://j3-fortran.org/doc/year/18/18-274.txt
* https://j3-fortran.org/doc/year/19/19-258.txt
* https://j3-fortran.org/doc/year/19/19-247r1.txt
* https://j3-fortran.org/doc/year/19/19-248.txt

Taking the second example from the `18-274` paper:
```fortran
  IF (PRESENT(D)) THEN
    CALL SUB(A,B,C,D)
  ELSE IF (X<1) THEN
    CALL SUB(A,B,C,EPSILON(X))
  ELSE
    CALL SUB(A,B,C,SPACING(X))
  END IF
```
One proposed syntax is "keyword syntax":
```fortran
CALL SUB(A, B, C, IF (PRESENT(D) THEN D ELSE IF (X < 1) THEN EPSILON(X) ELSE SPACING(X) END IF)
```
The second proposed syntax is "? syntax":
```fortran
CALL SUB(A, B, C, ? (PRESENT(D) D :? (X < 1) EPSILON(X) : SPACING(X) ?)
```

certik:
Which of these do people find the most readable and least readable?

My own feeling (from most readable to least readable):

1. Original
2. keyword syntax
3. ? syntax

We should also consider more use cases.

dev-zero:
To be honest I find both variants suboptimal (the first one would make an if/else/endif block return a value when being used inline, and simply run code as usual when not, but would be easier to read) and I wonder whether there is really a need for general conditional expressions or whether a ternary operator would be sufficient. In the latter case I would suggest the Python-based syntax (which one could even nest):

```fortran
CALL SUB(A, B, C, D IF PRESENT(D) ELSE (EPSILON(X) IF (X < 1) ELSE SPACING(X)))
```

certik:
@dev-zero in your opinion, can you rate by readability / preference the three options above plus your proposed Python-based syntax?

The ternary operator (that you presumably like) is the "? syntax" which is one of the proposed ideas, but you also say they are "suboptimal", so I am confused.

milancurcic:
From most to least readable:

1. Original
2. Python/@dev-zero syntax
3. Keyword syntax
4. ? syntax

Caveat: I use 2 in Python a lot so I'm used to it, thus my preference.

epagone:
FWIW my preference from most to least readable is:
1. Original
2. keyword syntax
3. ? syntax

[...]

PR #213, opened 2021-06-26
Initial draft of conditional expression intrinsic paper

certik:
The paper explains the motivation.

The other two referenced papers are:

* https://j3-fortran.org/doc/year/21/21-157.txt
* https://j3-fortran.org/doc/year/21/21-159.txt

@klausler, @milancurcic, would you have time to please review this and help me polish it? I would like to submit this for Monday's vote also.

### DIFF
--- a/proposals/conditional_expressions/cond.txt
+++ b/proposals/conditional_expressions/cond.txt
@@ -25,9 +25,9 @@ expressions. The number of `)` must match the number of `ifthen(` used.
 
 Examples:
 
-      ifthen(x >= 0.0, sqrt(x), -0.0)
+      res = ifthen(x >= 0.0, sqrt(x), -0.0)
 
-      ifthen(present(x), a, ifthen(present(b), b, 0))
+      res = ifthen(present(x), a, ifthen(present(b), b, 0))
 
 References:
 
--- a/proposals/conditional_expressions/cond.txt
+++ b/proposals/conditional_expressions/cond.txt
@@ -4,23 +4,21 @@ Subject: Conditional expressions using intrinsic function
 Date: 2021-June-26
 Reference: 21-157, 21-159
 
-Paper 21-157 proposes two syntactic forms for
-conditional expressions. The paper 21-159 proposes
-an alternative form. This paper proposes to implement
+Paper 21-157 proposes two syntactic forms for conditional expressions. The
+paper 21-159 proposes an alternative form. This paper proposes to implement
 conditional expressions using an intrinsic function.
 
-The syntactic forms in 21-157 are too verbose, as argued in 21-159. The form in
-21-159 is concise, but based on the poll [1], it is not the most popular
-either.
+The syntactic forms in 21-157 are too verbose, as argued in 21-159. The
+form in 21-159 is concise, but based on the poll [1], it is not the most
+popular either.
 
 We propose the following syntactic form
 
    <cond-expr> is ifthen( predicate, consequent, alternative )
 
-where the `predicate` is a scalar logical expression
-and the `consequent` and `alternative` are compatible
-expressions. The `alternative` can be another `ifthen` as shown in an example
-below.
+where the `predicate` is a scalar logical expression and the `consequent`
+and `alternative` are compatible expressions. The `alternative` can be
+another `ifthen` as shown in an example below.
 
 Examples:
 
@@ -32,8 +30,8 @@ Objections raised in the past:
 
 1) It is harder to do chaining due to the extra parenthesis at the end.
 
-Answer: it seems in practice it is not too bad, as in the second example above.
-Here is a longer chain:
+Answer: it seems in practice it is not too bad, as in the second example
+above.  Here is a longer chain:
 
       res = ifthen(present(x), x, ifthen(present(y), y,
                      ifthen(present(z), z, 0)))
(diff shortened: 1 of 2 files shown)


certik:
I believe this idea (not a specific proposal though) was discussed by the committee and the following feedback is summarized in:

* https://github.com/j3-fortran/fortran_proposals/issues/183#issuecomment-707843900
* https://github.com/j3-fortran/fortran_proposals/issues/183#issuecomment-703778043

Half the arguments are with issues that come up with extending `merge`, so we already avoid them by introducing a new intrinsic. The other half of the arguments we should address.

proposals/conditional_expressions/cond.txt
```diff
@@ -0,0 +1,34 @@
...
+
+The syntactics forms in 21-157 is too verbose, as argued in 21-159. The form in
+21-159 is concise, but based on the poll [1], it is not the most popular
+either.
+
+We propose the following syntactic form
+
+   <cond-expr> is ifthen( predicate, consequent,
+                      [ ifthen( predicate, consequent, ]*
+                      alternative) [)]*
```
milancurcic, line 20: I don't think this syntax is correct. In this syntax, only the first two arguments are required. But what is the result of:

```
res = ifthen(.false., 2)
```

I think the correct syntax is:

```suggestion
   <cond-expr> is ifthen( predicate, consequent, alternative )
```

and then one can use another call to `ifthen()` as the `alternative` if chaining is needed. Unless I'm missing some technical or semantic subtlety.

zjibben:
One other not-a-function aspect is that a conditional expression can return an actual variable, not just the value. So it can be used in `intent(out)` contexts like `allocate(x, stat=ifthen(logical, stat1, stat2))`. I don't imagine there's a precedent for that kind of behavior for a function. Still, the point stands that there's a precedent for function-looking things not really behaving like functions.

But your paper looks good and should provoke discussion.

milancurcic (commented):
Thanks, Ondrej! Besides my in-text comments, here are a few general comments and questions:

* Can this function be `simple`? I think so.
* In this approach, `consequent` and `alternative` must be type compatible. This is a restriction relative to the keyword (and other) forms, where you could do:
  ```
  res = if cond then 1 else 1.234
  ```
  or with a Python-like syntax:
  ```
  res = 1 if cond else 1.234
  ```
  Because of this restriction, I don't think `ifthen` could be used to invoke different specific procedures when passed as an argument to a generic procedure, which seems an attractive use case to me. But maybe I miss an important detail.

proposals/conditional_expressions/cond.txt
certik: I didn't know how to write it correctly, but the way it is written, `alternative` is out of `[]`, so it is not optional. So `ifthen(.false., 2)` is not allowed. The issue with `ifthen( predicate, consequent, alternative )` is that it might not be completely clear that in `ifthen( predicate, consequent, ifthen(predicate2, consequent2, alternative) )` the second `ifthen` is only evaluated if `predicate` is false. But I guess that follows from the general definition of `ifthen` that arguments are not evaluated unless needed and it applies recursively.

certik:
>     * In this approach, `consequent` and `alternative` must be type compatible. This is a restriction relative to the keyword (and other) forms, where you could do:
>       ```
>       res = if cond then 1 else 1.234
>       ```

First of all, the keyword version would be:
```
res = if (cond) then (1) else (1.234) endif
```
and my understanding is that it would be exactly equivalent to:
```
res = ifthen(cond, 1, 1.234)
```
The restriction on `consequent` and `alternative` would be exactly the same. So either both integers or both reals (depending on the type of `res`), and the default Fortran implicit casting rules apply.

certik:
> So it can be used in `intent(out)` contexts like `allocate(x, stat=ifthen(logical, stat1, stat2))`

Isn't this semantically exactly the same as:
```
allocate(x, stat=if (logical) then (stat1) else (stat2) endif)
```

I would think both could return a variable. But I haven't thought of it like this. I would expect both `if ... endif` and `ifthen` to always return a value.

milancurcic:
Yes, actually my second comment and examples are completely irrelevant, the keyword form has the same restriction for types. It wasn't obvious to me that it would have to have it, but it does simplify things.

proposals/conditional_expressions/cond.txt
milancurcic: You're right, somehow I didn't see `alternative` in your original form as I was writing the comment and example. What threw me off is that the optional arg comes before the required, which is not allowed. So if you put the `alternative` as the third argument, you don't need the thing in brackets.

certik:
I pushed in answers to some common objections. @milancurcic, @zjibben let me know if you have any other feedback, or if I should upload as is.

proposals/conditional_expressions/cond.txt
```diff
@@ -0,0 +1,54 @@
...
+                     ifthen(present(z), z, 0)))
+
+2) `ifthen` does not behave like a regular function, it is effectively new
+syntax, so we might as well actually introduce new syntax, to make it less
+confusing
+
+There is a precedent in optional arguments, where the argument is not evaluated
+but passed through through nested functions until it hits the function
+`present`. In a similar way, the arguments to `ifthen` are not evaluated ahead
+of time, but rather passes into the function `ifthen` itself, similar to how
```
milancurcic, line 48:
```suggestion
of time, but rather passed into the function `ifthen` itself, similar to how
```

milancurcic (approved):
I found two small typos, otherwise I think it's in good shape. The longer chained example is not bad at all, I find it quite readable. Thanks for leading this proposal.

certik:
Thanks @milancurcic for the quick review. I think @zjibben implicitly approved it, so I am going to go ahead and upload it.

certik:
Ok there it is: https://j3-fortran.org/doc/year/21/21-165.txt

ashe2:
What if you have this conforming program unit:
```
function f(cond, a, b)
  logical :: cond
  real :: f, a, b
  f = ifthen(cond, a, b)
end
```
Would this feature change the meaning of it? I.e. `ifthen` would change from an external to an intrinsic?

ashe2:
> > Would this feature change the meaning of it? I.e. `ifthen` would change from an external to an intrinsic?
> 
> Not if the program had a declaration of `ifthen`; a `use`, `interface`, `external`, &c would do.

As the example didn't have a declaration of `ifthen`, the meaning would change, right?

Is this a backward compatibility problem for any new intrinsic procedure?

klausler:
> > > Would this feature change the meaning of it? I.e. `ifthen` would change from an external to an intrinsic?
> > 
> > 
> > Not if the program had a declaration of `ifthen`; a `use`, `interface`, `external`, &c would do.
> 
> As the example didn't have a declaration of `ifthen`, the meaning would change, right?
> 
> Is this a backward compatibility problem for any new intrinsic procedure?

Yes, and yes.  This is always a consideration when standard or extension intrinsic functions are added.  If you use `IMPLICIT NONE(EXTERNAL)` , you're safe.

Issue #113, opened 2019-12-16
Augmented assignment


Libavius:
allow writing
```
a += b
```
instead of
```
a = a + b
```
and
```
a *= b
```
instead of
```
a = a * b
```

certik:
This is a very common request, so we should have an issue for it and document why such ideas were not implemented yet.

The last time I have seen this discussed is with this proposal: https://j3-fortran.org/doc/year/19/19-111r1.txt. It actually has reason why it was rejected.

Libavius:
> This is a very common request, so we should have an issue for it and document why such ideas were not implemented yet.
> 
> The last time I have seen this discussed is with this proposal: https://j3-fortran.org/doc/year/19/19-111r1.txt. It actually has reason why it was rejected.
> 
> (The `+=` and `*=` would also imply to have `/=`, but that is already taken for `.ne.`. The `19-111r1` paper thus introduces a new operator that could do the same.)

I think, just doing the two 'positive' operators would already help a lot. '-=' and '/=' are much less needed and can be circumvented like this:
The hypothetical code:
``` Fortran
a = 1.0_real64
do i=1, 10
    a -= vector(i)
end do
```
Can be written as:
``` Fortran
a = 1.0_real64
a *= -1
do i=1, 10
    a += vector(i)
end do
a *= -1
```
And analogous for division:
``` Fortran
a = 1.0_real64
a = 1/a; do i=1, 10
    a *= vector(i)
end do; a = 1/a
```
While this seems messy, it prevents the 'space ship operator' from being valid:
```
a -=- 1  ! this gives the same as a += 1
```
The main point however should be, that '-=' and '/=' are much less needed. I would really advise against another operator for '/='. That would be quite counter-intuitive.

PS: since the referred paper was in parts rejected due to a missing use case, I will ad one to the original post =)

certik:
Good point: the proposal would be to only add `+=` and `*=`. I would say the `+=` is probably the most common in practice. I grepped one of my codes for `i = i + 1` and I use just this particular string 7 times. I use `j = j + 1` 8 times, `k = k + 1` 3 times, etc., and that's just a small subset of all usages for `+=`. So having `+=` would simplify dozens of places in my own code for sure.

klausler:
The important aspect of BCPL's assignment operator(s) is that the left-hand side of the assignment is evaluated only once; e.g., if Fortran were to acquire such a feature, in `A(F()) += X`, there would only be one call made to `F`.

[...]

PR #238, opened 2021-11-01
Add augmented assignment proposal

Leonard-Reuter:
Added proposal for augmented assignment with `+.=`, `/.=` etc. following the discussion in #113

### DIFF
--- a/augmented_assignment/augmented_assignment.txt
+++ b/augmented_assignment/augmented_assignment.txt
@@ -1,0 +1,45 @@
+To: J3                                                     J3/XX-XXX
+From: Leonard Reuter 
+Subject: Augmented Assignment Operators
+Date: 2021-November-1
+
+Proposal for Fortran Standard: 202y
+
+
+1. Introduction
+
+The proposal is to introduce augmented assignment operators. Example:
+
+    i +.= 1
+
+Here, `i` is incremented by 1.
+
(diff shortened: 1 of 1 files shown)


certik:
Thanks for writing it up! 

In order for this proposal to have a chance, in my opinion we have to address existing and past committee criticisms.

You should reference previous proposals, I think there were some.

And then we need to go into the J3 mailinglist (I added a link in one of the comments in #113), and address every objection. Otherwise the same objections will be given and the proposal rejected.

augmented_assignment/augmented_assignment.txt
```diff
@@ -0,0 +1,45 @@
...
+
+    real :: A(n)
+    ...
+    A(get_index()) +.= x
+
+If the above example were to be written without augmented assigment, how
+often `get_index` is called would depend on the compiler.
+
+The dot is added between the operator and `=`, since `/=` is already in
+use.
```
ivan-pi, line 41: Here it would be good to add "in use as the *not equal* relational operator".

Leonard-Reuter:
@certik I updated the proposal by adding a discussion for the major objections.

@ivan-pi I added something similar to the proposal.

certik:
Thanks! Do you want me to merge it and do you want to submit it for the next committee meet?

Leonard-Reuter:
@certik Yes, that'd be great =) Thank you

Issue #87, opened 2019-11-15
Namespaces


certik:
@FortranFan wrote in https://github.com/j3-fortran/fortran_proposals/issues/86#issuecomment-554205627:

I have long debated in my own mind whether Fortran should formally introduce the concept of NAMESPACEs or whether MODULEs can themselves be elevated to first-class namespaces.  Looking at the proposal at #86 and with #1, I'm presently inclined toward the former.

Whilst I don't quite have a list of the requirements and specifications necessary toward an initial proposal, a sketch of the idea is to include either a new NAMESPACE scope and/or NAMESPACE attribute to MODULEs in addition to PUBLIC/PRIVATE visibility of MODULEs in a NAMESPACE.  Then a given NAMESPACE can IMPORT other NAMESPACEs.  Fortran can formally define a GLOBAL namespace and state all EXTERNAL MODULEs (a la external procedures) in a program are part of this namespace.  The language can then have some rules for name qualification and aliasing.  Code then might look like the following with illustrative syntax:

```fortran-modern
namespace my_lib
   public module constants
      real, parameter :: PI = 3.14
   end module
   private module utils !<-- can be USE'd only by modules in the same namespace
      interface
         module subroutine draw()
         end subroutine
      end interface
   end module
end namespace

namespace my_lib !<-- A namespace can be extended with more modules
   public module solver
      use constants, only : PI !<-- It can USE modules from the same namespace
   contains
      function calcradius( a ) result( r )
         real, intent(in) :: a
         result :: r
         r = sqrt(a/PI)
      end function
   end module
end namespace

namespace global !<- optional, for 'global' namespace will exist by default
   import my_lib, only : constants, solver ! import 'public' modules only from another namespace
   program p !<-- A program unit can only be part of GLOBAL namespace
      use constants, only : PI !<-- does not 'clash' with my_lib::constants
      real :: area
      print *, "PI = ", PI !<-- PI from USE-associated 'constants' module 
      area = 100.0
      associate ( PI => my_lib::constants%PI, calcrad => my_lib::solver%calcradius )
         print *, "circumference = ", 2.0*PI*calcrad(area)
      end associate
   end program
end namespace
```

Note the above is based on the use cases in #1 and #86; once more use cases are brought to light, it might help refine the idea of a NAMESPACE.

Libavius:
I can't decide yet, whether I prefer this over the 'module elevation'.

[...]

PR #262, opened 2022-05-13
Scoped enums

ZedThree:
This paper recommends that the new typed enumerators have values scoped to the type, and not to the enclosing scope.

> The authors _strongly_ believe that the current proposal for typed
> enumerators, as described in J3/21-120 and subsequent revisions, be
> amended so that enumerator names are class two names. That is,
> enumerator names must be local to each `enumeration type` and not in
> the surrounding scope. Some suggestions for syntax are made.

### DIFF
--- a/proposals/scoped_enumerators/scoped-enumerators.txt
+++ b/proposals/scoped_enumerators/scoped-enumerators.txt
@@ -1,0 +1,287 @@
+To: J3                                                     J3/XX-XXX
+From: Peter Hill
+Subject: Scoped Enumerators
+Date: 2022-05-13
+#Reference:
+
+Overview
+========
+
+The authors _strongly_ believe that the current proposal for typed
+enumerators, as described in J3/21-120 and subsequent revisions, be
+amended so that enumerator names are class two names. That is,
+enumerator names must be local to each `enumeration type` and not in
+the surrounding scope. Some suggestions for syntax are made.
+
+The rationale for making enumerator names class one (defined in the
+enclosing scope) is that this is "simpler". This is unfortunately too
+simple, and, unlike many of the other proposed and withdrawn features,
+if this becomes standardised it will be irrevocable. Extensible enums,
+arbitrary enumerator values, and underlying types could all be added
+to future standard versions, but unscoped enums are forever. Code
+written using unscoped typed enums would not be compatible with a
+future standard that introduced scoped typed enums (or worse, a third
+type of enum).
+
+Requiring enumerator names to be class two (defined local to the
+type), and the values accessed through some sort of namespace (for
+example, `<type> % <value>`), will provide benefits consistent with
+typed enums while still keeping open the possibility of a mechanism to
+"hoist" names into the enclosing scope (for example, see the C++
+"Using enum" paper P1099R5 which introduced this mechanism into C++20:
+https://atomgalaxy.github.io/using-enum/using-enum.html).
+
+As an additional benefit, introducing a syntax for
+type-namespace-access (for example the familiar component access `%`)
+leaves open the future possibility of a syntax for accessing
+`parameter` components or `nopass` type-bound procedures without the
+need for an instance of the type (usually called "static access" or
+similar in other programming languages).
+
+The rest of this paper will briefly review enums in other languages,
+the history of the typed enum proposal in Fortran, and the
+justification for changing to class two names.
+
+A brief review of `enum` in other programming languages
+=======================================================
+
+Enumerated types appear in many high-level programming languages, from
+Ada to Python. Of the most popular languages with a built-in `enum`
+facility, the majority introduce enumerator names within a unique
+scope for each enumeration type, so that referring to a particular
+enumerator value requires a namespace or class prefix.
+
+Languages which use global or enclosing scope:
+
+- Ada
+- C
+- Go
+- Pascal (some implementations enable local scopes instead)
+- Raku
+
+
+Languages which use local or own scope:
+
+- C#
+- C++ (with `enum class` in C++11)
+- D
+- Dart
+- Java
+- PHP
+- Python
+- Rust
+- Swift
+- Typescript
+
+
+Of the two above lists, it is notable that all of the languages that
+use a local scope are strongly typed, with the exception of PHP. While
+Ada, Go, and Pascal are also strongly typed, Ada provides a facility
+for resolving conflicting enumerator names, and the Free Pascal
+implementation supports scoped enums.
+
+Some of these languages go even further and make their enums be
+fully-fledged sum or algebraic types, where each enumerator may have a
+different type.
+
+Other languages, such as Javascript and CommonLisp, do not have a
+language-level enum but have other mechanisms for emulating them.
+
+Modern strongly typed languages that have enums have almost uniformly
+made them scoped enums, where enumerators must be accessed through the
+enum type. Some languages also provide mechanisms for abbreviated
+access: C++ has `using enum`, Swift allows dropping the type name in
+certain contexts where it is unambiguous.
+
+
+History and current status of scoped enumerations in Fortran
+============================================================
+
+`enum` was introduced in Fortran 2003 to enable interoperability
+between constants defined in Fortran and C. Similar to `enum` in C,
+the enumerators in Fortran's `enum` are defined in the enclosing
+scope.
+
+The following proposals (and their revisions) discuss enhanced `enum`s
+in some fashion:
+
+- 18-114: Enumeration types
+- 18-256: enums
+- 19-216: Enumeration types [US21]
+- 19-230: Formal Requirements for True Enumeration types
+- 19-231: Formal Requirements for True Enumeration types
+- 19-232: Formal Syntax for True Enumeration types
+- 19-239: Consider extended scope for enumeration types?
+- 19-249: Syntax for True Enumeration types
+- 21-110: Enumeration types discussion specs and syntax
+- 21-111: US-21 Typed enumerators
+- 21-120: Simpler enumeration types specs and syntax
+- 21-121: Enum types edits
+- 21-132: Simpler enumeration types, edits
+- 21-179: Enumerator accessibility and constructor
+- 21-182: US 21 Enumeration type.
+- 21-183: US 21 Enum type name.
+- 21-186: <enumeration-type-def> is not connected to high-level
+  syntax
+- 21-189: Interoperable enum types, additional specs/syntax/edits
+- 21-190: Editorial/technical fixes, mostly for enum types
+- 22-114: First Enumerator in an enumeration type should allow a
+  value.
+- 22-122: Enumeration type constructor
+
+The first paper, 18-114, suggests that enumerator names are introduced
+into the enclosing scope but provides a mechanism for resolving conflicts:
+
+> Example:
+>
+>   ENUMERATION, ORDERED :: COLORS
+>     ENUMERATOR :: WHITE, BROWN, GREEN
+>   END ENUMERATION COLORS
+>
+>   ENUMERATION, ORDERED :: NAMES
+>     ENUMERATOR :: GREEN, BROWN, WHITE
+>   END ENUMERATION NAMES
+>
+>   NAMES(GREEN) is a value of type NAMES.  COLORS(GREEN) is a value of
+>   type COLORS.  Their numeric representations are different.  GREEN
+>   alone is prohibited because its type is ambiguous.
+
+19-230r2 has a looser requirement:
+
+> F.  The name of an enumerator must be unique within the enumeration type,
+>     but may be the same as the name of an enumerator of another type.
+>     Some mechanism shall allow access to both enumerators from a
+>     scope.
+
+19-231r2 goes further and recommends scoped enumerators:
+
+> 13. [F]   An enumerator is a constant expression.
+>
+> 14. [F]   Accessing the name of an enumeration type by use association does
+>           not make any of its enumerator names available (because
+>           enumerator names are not class (1) names).
+>
+> 15. [F]   The syntax for denoting an enumerator shall involve the
+>           enumeration type name and no other class (1) name. If the
+>           enumeration type name is inaccessible, this syntax shall not
+>           be available.
+
+19-232 suggests some syntax:
+
+> (e) Syntax for "qualified denotation", i.e. specifying an enumerator within
+>     a specific named enumeration type. As the enumerators live inside the
+>     type, just use the type name followed by the type-access token %
+>     and the enumerator name, e.g.
+>         MY_ENUM % HIDDEN_ENUMERATOR_NAME
+>
+> ALTERNATIVE SYNTAX:
+> (e1) Use a new token instead of reusing %; e.g.
+>         MY_ENUM ` HIDDEN_ENUMERATOR_NAME
+>
+> (e2) Use the type constructor with the inaccessible enumerator name as its
+>      argument, e.g.
+>         MY_ENUM ( HIDDEN_ENUMERATOR_NAME )
+>
+> COMMENT: e2 syntax is ambiguous if MY_ENUM is to be treated as a generic
+>          function, as we do with derived types since Fortran 2003. Even if
+>          there is no ambiguity in a particular example, it promulgates
+>          confusion between local entity names and names inside a type.
+
+The withdrawn paper 19-249 notes that, in regard to the `MY_ENUM %
+HIDDEN_ENUMERATOR_NAME` syntax:
+
+> {We do not ever use "<type-name> %" for anything.}
+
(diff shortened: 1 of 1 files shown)


certik:
@ZedThree thanks a lot for writing this up! I will read it carefully soon and provide feedback.

Let's iterate on this. Everybody, go ahead and provide feedback as well please.

klausler:
I completely agree that the current 202X typed enumerators are ridiculous and should be deleted or replaced.  Thank you for writing this up.  There is zero chance that J3 will fix this, but maybe we implementers can agree on a better common solution.

Three suggestions:

(1) The syntax "T%E" is ambiguous at parsing time with a component reference.  Consider another symbol, please.  "T:E" would work.  So would "T_E" if you want really easy implementation.

(2) The class-1 scoping of the current 202X misfeature could be retained with a keyword, like "ENUM, OPEN" or something.

(3) We have to also have optional value initializers for interoperability.

If an unambiguous syntax is used for literal constant enumerators, all of this would be trivial to implement in f18.  A syntax that's ambiguous with component references could be made to work -- misparses one way or the other would have to be detected and repaired once the symbol table is in hand, as I have to do for statement function definitions vs array element assignments, function calls vs array references, array slices vs substrings, &c. &c. -- but would be a harder sell to me if anybody wanted a demonstrable prototype in short order.

ZedThree:
@klausler Thanks for your comments, it's good to hear from an implementer!

(1) The appeal of `%` is that (to me at least) it feels quite intuitive what it means, but if it's impediment to implementation, that's certainly a strike against it. `:` feels quite overloaded already, maybe `::`? Or even the ugly `%%`? Early revisions suggested `enum_name(value_name)` as a way to resolve conflicts. Could that be made to work?

Do you think if this had a straw poll for syntax, it would help?

(2) Indeed, or perhaps `import` which is already used for similar purposes in interfaces. I'm not proposing that feature here though, as I thought this might stand a better chance by keeping the scope (ha) limited.

(3) I'm not sure what you mean -- could you elaborate?

certik:
(1) The syntax of accessing the enumerators `T%E` should be designed to be consistent with name spaces for modules (#1), which have a similar issue: `some_module%some_function()`. As well as if Fortran was to get namespaces (#87), they should use a consistent syntax. As an example, C++ uses `.` for "derived types", but `::` for namespaces and enumerations (in C++ it seems the enumerators are both global as well as accessible locally using `::` for `enum`, but only locally for `enum class`). So if we got with `T:E`, then also `some_module:some_function()` should work. What about `T::E`, `some_module::some_function()`?

plevold:
I think `T::E` makes the most sense both for namespaces and for enums (and `nopass` procedures for that matter). This is common in multiple similar programming languages. Insisting on that Fortran syntax should be "unique" only makes it confusing for newcomers. As a comparison, Rust is considered to be "the C++ replacement" and has adopted much of its syntax like `::` for namspaces and `<>` for generics.

@ZedThree It might be worth pointing out in the paper that with the current Fortran standard it's trivial to implement a strongly typed and properly namespaced enum. I've often done it in my code like the example below. It should be easy to reduce the amount of boilerplate code with e.g. fypp or maybe even compiler preprocessors as well. This begs the question if an arguably worse implementation is even needed in the standard.

```fortran
module animal_mod
    implicit none

    private
    public animal_t, animals

    type :: animal_t
        private
        integer :: id
    contains
        generic :: operator(==) => equals
        generic :: operator(/=) => not_equals

        procedure, private :: equals, not_equals
    end type

    type :: animals_t
        type(animal_t) :: dog = animal_t(1)
        type(animal_t) :: cat = animal_t(2)
        type(animal_t) :: bird = animal_t(3)
        type(animal_t) :: fox = animal_t(4)
    end type

    type(animals_t), parameter :: animals = animals_t()

contains

    logical elemental function equals(a,b)
        class(animal_t), intent(in) :: a
        type(animal_t), intent(in) :: b
        equals = a%id == b%id
    end function equals


    logical elemental function not_equals(a,b)
        class(animal_t), intent(in) :: a
        type(animal_t), intent(in) :: b
        not_equals = .not. a == b
    end function not_equals

end module

program  main
    use animal_mod, only: animal_t, animals
    implicit none

    write(*,*) what_says(animals%dog)
    write(*,*) what_says(animals%cat)
    write(*,*) what_says(animals%bird)
    write(*,*) what_says(animals%fox)

contains

    pure function what_says(animal) result(sound)
        type(animal_t), intent(in) :: animal
        character(len=:), allocatable :: sound

        if (animal == animals%dog) then
            sound = 'woof'
        else if (animal == animals%cat) then
            sound = 'meow'
        else if (animal == animals%bird) then
            sound = 'tweet'
        else
            sound = '???'
        end if
    end function

end program
```

Program output:
```
 woof
 meow
 tweet
 ???
```

FortranFan:
> it's trivial to implement a strongly typed and properly namespaced enum. I've often done it in my code like the example below. It should be easy to reduce the amount of boilerplate code with e.g. fypp or maybe even compiler preprocessors as well. This begs the question if an arguably worse implementation is even needed in the standard.

An important need with proper enums is use with `SELECT CASE` constructs that no "roll your own" types permit.

FortranFan:
> Let's iterate on this. Everybody, go ahead and provide feedback as well please.

Honestly, @certik , at this stage the "ball" is in WG5's side of the "court" entirely, the current design was passed by J3 with no objections by any of the voting members.  You can be completely certain J3's position will be "There will be no further action on these suggestions at this time."

The only viable option is to be submit a request with WG5 to withdraw the `enumeration type` feature from Fortran 202X.

FortranFan:
> This paper recommends that the new typed enumerators have values scoped to the type, and not to the enclosing scope.
> ..

@ZedThree , there are few use cases which are supported by the 'enumeration type` as drafted into Fortran 202X.

In my opinion, to pick one use case that is of interest to the authors and proposing something to get that addressed is **just not right.**  I can't give a big enough down vote on this, it is thumbing its nose at all the other use cases.

certik:
@FortranFan, can you formulate what you propose should be done? Do you like the enumerations as proposed in 2X more than this proposal?

FortranFan:
> @FortranFan, can you formulate what you propose should be done?

The only viable option is to submit a request with WG5 to withdraw the enumeration type feature from Fortran 202X.

klausler:
Is there any essential difference between this specific proposal -- putting enumerators into a scope and requiring some kind of syntax that names the scope and the enumerator wherever an enumerator appears -- and using distinct prefixes on the enumerators?  It seems to me that we're just saving some typing here.

certik:
If you use distinct/unique prefixes (and are willing to accept this downside) then I think there is no difference.

septcolor:
I am not familiar with enums (so now learning how to use it), but I think qualifiers are useful for locally renaming the prefix of enumerators as a whole, as needed (e.g. to avoid name collisions or for code brevity), rather than renaming each enumerator separately.
```
#include <iostream>
namespace NS1 {
    enum class Color { red, green, orange, blue };
}
namespace NS2 {
    enum class Color { red, green, blue };
}

int main() {
    // These may correspond to "use NS1, only: Col1 => Color" etc.
    using Col1 = NS1::Color;
    using Col2 = NS2::Color;
    std::cout << (int) Col1::blue << std::endl;
    std::cout << (int) Col2::blue << std::endl;
}
```

ZedThree:
@FortranFan :
> @ZedThree , there are few use cases which are supported by the 'enumeration type` as currently included in draft Fortran 202X.
> 
> In my opinion, to pick one missing use case that is of interest to the authors and proposing something to get that addressed is **just not right.** I can't give a big enough down vote on this, it is thumbing its nose at all the other use cases.

I disagree, the existing proposal supports most use cases even if it is missing several features that would support more use cases, the main two being the ability to specify the underlying type, and values for individual enumerators. Both of these could be added in F202Y without breaking code that uses the current spec, and could be vendor extensions until then. I certainly disagree that this is "thumbing its nose at other use cases" -- this proposal is about enhancing all use cases, and not about any particular use case.

The existing spec certainly has its flaws and the final form seems to have ignored a couple of decades of language research and innovation, but I don't think it is so flawed that it should be completely thrown out.


@klausler :
> Is there any essential difference between this specific proposal -- putting enumerators into a scope and requiring some kind of syntax that names the scope and the enumerator wherever an enumerator appears -- and using distinct prefixes on the enumerators? It seems to me that we're just saving some typing here.

It's not _just_ about saving typing, although that is part of it, it's about building in best coding practice. Rather than having to teach new developers "you should always add a unique prefix to enum names, because sometimes it's important, and if you don't you might have to rename each individual enumerator when you `use` them", we can just teach them how to use them and that's enough.

More concretely, here's an example:

```Fortran
! Existing spec using renaming:
use colour_mod, only: colour_t, colour_red => red, colour_green => green, colour_blue => blue
use alert_mod, only: alert_t, alert_red => red, alert_amber => amber, alert_green => green
! or with prefixes:
use colour_mod, only: colour_t, colour_red, colour_green, colour_blue
use alert_mod, only: alert_t, alert_red, alert_amber, alert_green

! This proposal:
use colour_mod, only: colour_t
use alert_mod, only: alert_t
```

plevold:
> An important need with proper enums is use with `SELECT CASE` constructs that no "roll your own" types permit.

@FortranFan that's true, but with the proposed `select case` statement my understanding is that you will only save a few keystrokes. Sure, the code

```fortran
if (animal == dog) then
    sound = 'woof'
else if (animal == cat) then
    sound = 'meow'
else if (animal == bird) then
    sound = 'tweet'
else
    sound = '???'
end if
```

could be rewritten to 

```fortran
select case(animal)
case (dog)
    sound = 'woof'
case (cat)
    sound = 'meow'
case (bird)
    sound = 'tweet'
case default
    sound = '???'
end select
```
That should give you 114 non-whitespace characters instead of 123 and maybe somewhat better readability. The behavior is exactly the same though. 

On the other hand, if the Fortran compiler would require that the `select case` statement was exhaustive, either by including all enum variants or by having a `case default` statement then there would be an actual advantage of using `select case` over `if`-s.

>Is there any essential difference between this specific proposal -- putting enumerators into a scope and requiring some kind of syntax that names the scope and the enumerator wherever an enumerator appears -- and using distinct prefixes on the enumerators? It seems to me that we're just saving some typing here.

@klausler in addition to the example by @ZedThree another advantage of using scope instead of a prefix is that (a future) Fortran standard could allow the scope to be omitted in `case (...)`-statements since the enum type is already known from the `select case (var)` line. For example, 

```fortran
select case(animal)
case (animal_t::dog)
    sound = 'woof'
case (animal_t::cat)
    sound = 'meow'
case (animal_t::bird)
    sound = 'tweet'
case default
    sound = '???'
end select
```

could be simplified to

```fortran
select case(animal)
case (dog)
    sound = 'woof'
case (cat)
    sound = 'meow'
case (bird)
    sound = 'tweet'
case default
    sound = '???'
end select
```
still without the risk of naming collisions. This is for example allowed in Java switch-statements.

klausler:
If you're just saving typing, it's not worth the effort and drama.  Add int values and I/O and you'd have something.

`SELECT CASE` is O(1) or O(log2(N)).  An `IF` cascade is O(N).  More important, a `SELECT CASE` with no default can be checked at compilation time for missing values.

plevold:
@klausler good point about the performance differences. So with the current proposal one would have to choose between performance (with `enumeration type`) or maintainability (with a custom implementation). In my experience, maintainability trumps performance in large software projects unless the penalty is absolutely prohibiting. 

> More important, a SELECT CASE with no default can be checked at compilation time for missing values.
Absolutely agree. My understanding is that with the current 202X proposal compilers are not required to give an error on missing values, but maybe I've misread something?

ZedThree:
As far as I can see it, these are the missing features that one would want from an enum in a modern language:

- proper scope
- exhaustive `select case`
- custom values
- custom underlying type
- proper I/O using the enumerator name (reflection)\*

Of these, I think the first two cannot be implemented in a future standard without breaking backwards-compatibility, while the other three can. Are there others? I'm ignoring proper sum/algebraic types here, because I think they would only be usable with proper generic typing support.

Several people that have knowledge of how the committee works have said this proposal doesn't stand any chance of getting accepted. Is "kill/delay typed enumerations" more likely to get accepted instead?

[21-110](https://j3-fortran.org/doc/year/21/21-110.txt) is the paper that stripped down the much nicer [19-216](https://j3-fortran.org/doc/year/19/19-216r1.txt). Does anybody have any insight into what the committee thought was too complicated about 19-216?

---
\* Weirdly, the current spec has some degree of reflection with `next()/previous()` and `enum_type(1)/huge(enum_type)`, not seen in many other languages

klausler:
> Of these, I think the first two cannot be implemented in a future standard without breaking backwards-compatibility, while the other three can. Are there others? I'm ignoring proper sum/algebraic types here, because I think they would only be usable with proper generic typing support.

Proper sum types as in ML/O'Caml/Hashkell would want their constructors (including nullary ones) to be available in the declaration scope anyway.

> Several people that have knowledge of how the committee works have said this proposal doesn't stand any chance of getting accepted. Is "kill/delay typed enumerations" more likely to get accepted instead?

No.  There's huge overlap between WG5 and J3.  J3 asking WG5 for permission to defer the feature involves the same group of people taking off their J3 hats, donning their WG5 hats, and saying "no" to themselves.

ZedThree:
> No. There's huge overlap between WG5 and J3. J3 asking WG5 for permission to defer the feature involves the same group of people taking off their J3 hats, donning their WG5 hats, and saying "no" to themselves

I'm sorry, I am entirely ignorant of the structure here. There's no chance of changing J3's mind on this then either?

certik:
@ZedThree there is always a chance. That's why I asked you to write this up. Even if they say no, we can require the committee to provide a response why the paper was not accepted. Furthermore, it provides guidance to compiler developers to implement as extensions.

There is committee politics and how it operates, but I urge all participants here to stay away from that, as it is not productive (or move it to a separate "meta" issue). In this issue, let's focus on having free discussion about what makes sense to do for enumerations.

Peter K. has good points about it being "not worth doing", as it is "just syntax". But, isn't most of Fortran "just nice syntax'? It absolutely is, and that is why I like it. And I think having modern enumerations in the language should be the goal, and we should not be discouraged. Peter H.'s arguments thus resonate with me. Finally, I think @FortranFan's main argument is that this proposal doesn't go far enough, and thus is not worth it, but I think it's a step in the right direction and most importantly: it allows to add the other things in backwards compatible manner. If this proposal is not accepted, and the current F2023 enumerations go in, then I think we can't change the scope without breaking backwards compatibility.

klausler:
> > No. There's huge overlap between WG5 and J3. J3 asking WG5 for permission to defer the feature involves the same group of people taking off their J3 hats, donning their WG5 hats, and saying "no" to themselves
> 
> I'm sorry, I am entirely ignorant of the structure here. There's no chance of changing J3's mind on this then either?

WG5 is the ISO (international) standard committee for Fortran.  J3 is the old name for the American standard committee for Fortran.  WG5 comes up with lists of work items for each new standard, and asks J3 to define them; J3 writes draft international standards, gives them to WG5, and they get approved as ISO standards.  J3 also takes care of corrections and interpretation requests.

One problem with this arrangement is that it has no feedback loop that includes implementation experience or even broad review before WG5 hands ideas off to J3, and in the 3+ decades since the structure was established, there's a sad history of not-completely-thought-through changes being set permanently into the language.  That's why we have PDTs still not being implemented widely enough to be portable after 20 years; a `DO CONCURRENT` construct that is serial; and soon, rank-agnostic array indexing.  And we don't have simple things like pointers to immutable data.  It is a mess.

plevold:
> I'm ignoring proper sum/algebraic types here, because I think they would only be usable with proper generic typing support.

I think proper sum types has a lot use cases without proper generic types (and even more with it). How JSON values are parsed in Rust is a good practical example: When reading data at a given path, e.g. `root["foo"]["bar"]` you get back a [Value](https://docs.serde.rs/serde_json/value/enum.Value.html) enum:

```rust
pub enum Value {
    Null,
    Bool(bool),
    Number(Number),
    String(String),
    Array(Vec<Value>),
    Object(Map<String, Value>),
}
```
While the data of the `Array` and `Object` variants use generics this enum doesn't. If Fortran enumration types had support for different types it would be possible to implement this kind of interface in Fortran as well without any additional language features.

For reference, the alternative "object oriented" approach (one that is possible to implement in Fortran today) would be to use an abstract base class:

```fortran
type, abstract :: json_value_t
end type

type, extends(json_value_t) :: json_bool_t
    logical :: value
end type

type, extends(json_value_t) :: json_number_t
    real(dp) :: value
end type
```

And so on. This approach is (of course) very verbose and does not provide much type safety: While you could `select type` on a `class(json_value_t)` variable there's no way of knowing that you've covered all possible variations since inheritance is not an finite set . Even a type outside this hypothetical JSON parsing library could extend `json_value_t`.

PR #291, opened 2023-01-12
proposal to add missing IEEE-754 functions

### LINKED ISSUE #222: `log2` instrinsic
Currently the standard has `log` and `log10` functions, but no `log2`.

everythingfunctional:
This proposes to add all of the missing functions from IEEE-754 Table 9.1.

Note: I did not add `exp2` or `exp10` as they are expressible as `2**x` and `10**x` already, but I would be willing to add them if there are some who believe there is reason to do so. Ditto for `pow` and `pown`, but I believe there is even less chance of reason to add them since `x**y` and `x**n` are exactly the same.

Fix #222

### DIFF
--- a/proposals/math_functions/ieee-754-math-functions.txt
+++ b/proposals/math_functions/ieee-754-math-functions.txt
@@ -1,0 +1,292 @@
+To: J3                                                     J3/##-###
+From: Brad Richardson
+Subject: IEEE-754 Recommended Math Functions
+Date: 11-Jan-2023
+
+#Reference:
+
+Introduction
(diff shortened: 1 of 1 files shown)


everythingfunctional:
Note to those interested, IEEE-754 2019 can be found [here](https://ieeexplore.ieee.org/document/8766229).

bonachea:
Superseded by:
* [26-123r4](https://j3-fortran.org/doc/year/26/26-123r4.txt) US19 - Missing IEEE functions, Edits

Issue #323, opened 2023-12-21
Complex pointer to a real array and vice-versa


PierUgit:
In many codes that use FFTs and the Fourier domain one need to alternatively work on the same array as real or as complex numbers. Different tricks have to used, none being really satisfactory, e.g.:
```
real, allocatable :: r(:)
! ...
! some computations on r(:)...
call rfft(r) ! real to complex in-place transform
call compute_complex(r,size(r))   ! this routine MUST NOT have any interface !!!
call irfft(r) ! complex to real in-place inverse transform
! ...

subroutine(c,n)
complex :: c(n/2)
! some computations on c(:)
end subroutine
```

Instead, the standard should provide a way to have a complex pointer to a real array:
```
real, allocatable, target :: r(:)
complex, pointer :: c(:)
! ...
! some computations on r(:)...
call rfft(r) ! real to complex in-place transform
c => cmplx_pointer(r)
! some computations on c(:)...
call irfft(r) ! complex to real in-place inverse transform
! ...
```
And vice-versa 
```
complex, allocatable, target :: c(:)
real, pointer :: r(:)
! ...
! r => real_pointer(c)
! some computations on r(:)...
call rfft(r) ! real to complex in-place transform
! some computations on c(:)...
call irfft(r) ! complex to real in-place inverse transform
! ...
```
Of course some constraints would be needed:
- the size of the real array in the first dimension shall be even
- the real or complex array shall have a unit stride along the first dimension (forcing the whole array to be contiguous could be desirable, so that the compiler could catch the error at compile-time)
- `c(:)%re` shall be equivalent to `r(1::2)` and `c(:)%im` shall be equivalent to `r(2::2)`, which enforces the real component to be stored first (de-facto standard in all compilers that I know).

certik:
One thing that might be relevant here: often times in FFT it is possible to compute the transformation faster if all the real parts are stored first, and then all the imaginary parts. In other words, if `c(:)%re` is equivalent to `r(1:n/2)` and `c(:)%im` is equivalent to `r(n/2+1:n)`. It would be cool if Fortran compilers allowed this alternative layout of complex arrays. It seems the language itself almost allows it, except some corner cases.

PierUgit:
@certik This goes beyond and looks more complex than my proposal, which is actually very simple. 

An alternative syntax has been proposed on the discourse group:
`c => complex :: r(:)`

PierUgit:
[...]

PR #325, opened 2024-01-03
Adding a proposal to allow complex pointers to real arrays and vice-versa

PierUgit:
Following the discussion in #323

### DIFF
--- a/proposals/cmplx-real-pointers.txt
+++ b/proposals/cmplx-real-pointers.txt
@@ -78,11 +78,11 @@ vice-versa. The above code would become:
 No new syntax need to be created, the usual pointer association can be
 used:
 
-`c => complex_pointer(r)`
+`c => r`
 
 - `r` is a *contiguous* real array, which has either the target or the
   pointer attribute
-- `c` is a complex array pointer of the same rank than `r`
+- `c` is a complex array pointer of the same rank and same kind than `r`
   - `c` can also be a complex scalar pointer, in the case where r is a
     rank-1 array of size 2
 - the size of the first dimension of `r` shall be even
@@ -91,12 +91,12 @@ used:
 - `c%im` refers to the same storage than `r(2::2)` (rank-1), or 
   `r(2::2,:)` (rank-2), etc...
 
-`r => real_pointer(c)`
+`r => c`
 
 - the exact opposite
 - `c` is a *contiguous* complex array or a complex scalar, which has 
   either the target or the pointer attribute
-- `r` is a complex array pointer of the same rank than `c`
+- `r` is a complex array pointer of the same rank and same kind than `c`
   - if `c` is a scalar, then `r` is a rank-1 pointer of size 2
 - same other rules as above
 
--- a/proposals/cmplx-real-pointers.txt
+++ b/proposals/cmplx-real-pointers.txt
@@ -10,6 +10,7 @@ The proposal is to allow a complex pointer to be associated to a real
 array, and similarly to allow a real pointer to be associated to a 
 complex scalar or array. 
 
+
 2. Motivation
 
 In short, the motivation is to save memory and cpu-memory bandwidth
@@ -23,12 +24,13 @@ complex numbers in the same real array, with interleaved real and
 imaginary parts. Once in the Fourier domain the complex type is more
 appropriate to perform computations, but this requires either:
 - duplicating the data into a complex array after the transform, 
-  thus cancelling the advantage of the in-place transform. Note that in HPC,
-  and particularly when dealing with large data volumes, duplicating the
-  data is really something that we want to avoid whenever possible.
-- or using various non-standard tricks that rely on the fact that virtually 
-  all compilers store a complex number by the sequence real part/
-  imaginary part.
+  thus cancelling the advantage of the in-place transform. Note that in 
+  HPC, and particularly when dealing with large data volumes, 
+  duplicating the data is really something that we want to avoid 
+  whenever possible.
+- or using various non-standard tricks that rely on the fact that 
+  virtually all compilers store a complex number by the sequence 
+  real part / imaginary part.
 
 For instance, consider a code that performs a FFT based convolution
 at some point:
@@ -83,7 +85,8 @@ library [1].
 
 More generally, this is a recurrent topic on forums or on 
 StackOverflow [2]
-	
+
+
 3. Proposed solution
 
 The proposal essentially aims at standardizing the usual practice 
@@ -163,7 +166,28 @@ call storage_associate(r,c[,shape][,lower])
 call storage_associate(r,c[,shape][,lower])
 ```
 
-3.3 Prototyping, etc...
+3.3 Intermediate representation
+
+During the discussions it has been proposed tu use an intermediate
+representation between the real and complex view, which would be a real
+view with an augmented rank and first dimension of size 2. Instead of
+going directly from `r(n)` to `c(n/2)` one would define by rank
+remapping:
+`r2(1:2,1:n/2) => r(1:n)`, then `c => r2`
+And the other way:
+`r2 => c`, then `r(1:n) => r2(1:2,1:n/2)`
+
+This could also be coupled with a new pseudo-component notation `c%reim`
+that would be equivalent the `r2` respresentation (i.e. `c%reim` would be
+`real`, with a shape [2,n/2]). 
+
+I personally find that such an intermediate representation just makes 
+the proposal more complex than it needs to be, and that nothing really 
+useful can be performed with it. Moreover the `c%reim` notation would 
+require another change in the standard to allow non zero ranks on both 
+sides of `%`.
+
+3.4 Prototyping, etc...
 
 I think that this proposal doesn't need to have a preliminary prototype 
 implementation, as it essentially consists in standardizing an already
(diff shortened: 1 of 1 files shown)


proposals/cmplx-real-pointers.txt
```diff
@@ -0,0 +1,139 @@
...
+
+4. Objections
+
+4.1. Memory layout
+
+This proposal would constraint the memory layout of the complex type 
+beyond what the current standard does. However, the required layout 
+seems to be the de-facto standard, used by virtually all existing 
+compilers, and it's hard to think about a possible drawback of 
+constraining the layout like this.
```
certik, line 122: The drawback of constraining the layout is that we can't represent the "split" (block) layout natively, but the split layout might have better performance. (We can't do the split layout today, but this might be a good extension in the future.)

Here is a page in FFTW that talks about this: https://fftw.org/fftw3_doc/Interleaved-and-split-arrays.html.

Here is another: https://developer.apple.com/documentation/accelerate/data_packing_for_fourier_transforms?language=objc

It looks like the terms for this are:

* Interleaved-Complex Format
* Split-Complex Format

The advantage of the spit-complex format is performance (and Fortran cares about performance). Here is a page that talks about it: https://www.jobilize.com/course/section/split-format-vs-interleaved-format-by-openstax, citing:

> The benefit of operating in split format is obvious: the shuffle operations that were required in [[link]](https://www.jobilize.com/online/course/0-2-implementation-details-computing-the-fast-fourier-transform-by-ope?page=6#uid20) are avoided because the real and imaginary parts can be implicitly swapped at the instruction level, rather than by awkwardly manipulating SIMD registers at the data level of abstraction.

Roughly speaking, the interleaved-complex format requires a shuffle (swap) SIMD instruction for many complex operations (such as multiply) that the split-complex format doesn't.


Coming to your proposal: one way to resolve it is that for interleaved-complex format you would do:
```
c1 => complex_pointer(r1)
```
And for split-complex format you would do:
```
c1 => complex_pointer(r1_real, r1_imag)
```
Where the `c1` complex array is declared with the appropriate `layout(interleaved)` and `layout(split)` layouts.

As long as we keep this in mind (to not close the door to the split format), then your proposal should work.

PierUgit: As discussed in the issue, I have some serious doubts about the possibility of introducing a complex split layout in the standard, as it would result in scalars that are possibly stored non contiguously. Assuming it's possible, I think that the proposal doesn't really close the door:
> the obtained view would be layout dependent, and the standard would have to describe the view for each layout
In the case of a split layout, `c%re` would refer to the same storage as `r(1:n/2)`

I can make it more explicit in the proposal.

certik: Excellent, your commit https://github.com/j3-fortran/fortran_proposals/pull/325/commits/a928885dce400f26f5fbd08c0556ba9868344403 addresses my comment, thank you. Now it's good I think.

certik:
I think this is a good proposal now. You can ask at the Fortran Discourse to get more feedback.

everythingfunctional:
IMO, the additional intrinsics and/or new syntax are superfluous. The storage sequence is already sufficiently defined (or is nearly trivial to add clarification in case it's not) that all that's needed is to say that a complex pointer may be associated with a real target of the same kind and vice-versa. All the other existing constraints and semantics should take care of everything else, barring the edge case of ensuring that the size of the real target must be even.

PierUgit:
> IMO, the additional intrinsics and/or new syntax are superfluous.

Maybe. I wanted such association to be explicit, in order to avoid possible mistakes, but it can also makes sense to stick to the usual syntax.

PierUgit:
Actually, for full consistency, I also think that the associating a real actual argument to a complex dummy argument, and vice-versa, should be allowed, with the same rules and restrictions as for the pointer association.

ivan-pi:
Perhaps it's worth referencing the section 6.2.5, paragraph 17, page 39, from the draft C23 standard N3096 (https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3096.pdf),

> Each complex type has the same representation and alignment requirements as an array type
containing exactly two elements of the corresponding real type; the first element is equal to the real part, and the second element to the imaginary part, of the complex number

This would imply that the interoperable Fortran type `complex(c_double)` (or `complex(c_double_complex)`) and `real(c_double)` also share the same representation and alignment requirements, meaning that storage association is possible.

PierUgit:
@ivan-pi I actually think that this would be implied to make the association work expected.

everythingfunctional:
> I also think that associating a real actual argument to a complex dummy argument, and vice-versa, should be allowed, with the same rules and restrictions as for the pointer association.

Actually, you don't want to be able to do this, except perhaps for assumed size or explicit shape (which imply the actual must be contiguous), or explicitly contiguous. I.e. what happens in the following case?

```fortran
r = [1., 2., 3., 4.]
call sub(r(1:3:2))
print *, r
contains
subroutine sub(c)
  complex, intent(inout) :: c(:)
  print *, c(1)%im, c(1)%re
  c(1)%im = 42.
end subroutine
end
```

or even crazier

```fortran
r = [1., 2., 3., 4.]
call sub(r([4,1]))
print *, r
contains
subroutine sub(c)
  complex, intent(inout) :: c(:)
  print *, c(1)%im, c(1)%re
  c(1)%im = 42.
end subroutine
end
```

IMO, better not to get into all the complexities to define argument association, as you can still do it (albeit with more lines) with the pointer association.

PierUgit:
@everythingfunctional It doesn't look so bad to me. 

In your first example the compiler would detect a real non-contiguous actual argument vs a complex dummy argument, and it would apply the contiguity rule in this case: either it reports a compilation error, or it generates copy-in/copy-out. It's actually not very from `c => r(1:3:2)` (not allowed because the RHS is non contiguous)

And the same for the second example (actually this one wouldn't be valid even with a real dummy argument, as an array-indexed actual argument can not be passed to an intent(out) dummy argument IIRC.

PierUgit:
I meant that the compiler can check the contiguity (as well as other requirement) equally well for the argument association than for the pointer association.

PierUgit:
I have update the proposal with the optional extension to argument association.

proposals/cmplx-real-pointers.txt
```diff
@@ -76,13 +77,14 @@ vice-versa. The above code would become:
...
-used:
+used, with additional rules and restrictions:
 
 `c => r`
 
 - `r` is a *contiguous* real array, which has either the target or the
   pointer attribute
-- `c` is a complex array pointer of the same rank and same kind than `r`
+- `c` is a complex array pointer of same kind than `r`, and of of the
+rank by default (but pointer rank remapping can be used)
```
everythingfunctional, line 87:
```suggestion
- `c` is a complex array pointer of the same kind as `r`, and of the same
rank by default (but pointer rank remapping can be used)
```

aradi:
It would be probably worth to consider to remove also the direct pointer association `r => c` and leave only the one achieved via the call of an intrinsic subroutine. This way, the changes to the standard would be very well localized. The "type breaking" would only occur at one place.

It would be probably also worth to make the subroutine name more generic (`e.g. call associate_storage(r, c, [shape], [stat])`), so that in the future it can also work for other type combinations, if they make sense (e.g. array of bytes/characters associated with integers, etc.) And I'd suggest to add a `stat` optional argument, which can report back, if the storage association was not successful (e.g. types are not compatible, size of the target is incommensurable with the pointer type, etc.)

PierUgit:
What's next with this proposal? What should I/we do now?

certik:
@PierUgit now you should attend a J3 meeting (at least remotely) and advocate for it. If you can't do that, find a committee member who can do that for you.

everythingfunctional:
It will go a long way if you commit to helping do the work on this all the way through the process. I.e.

1. Use cases (which you've got a decent start on here)
2. Requirements
3. Specification
4. Syntax
5. Edits to the standard

I'm willing to upload papers and facilitate discussion in plenary on someone's behalf if you can't join the committee or attend the meetings, but I'm not going to advocate other members spend time on features I'm not passionate about. And I don't really have time to work on more than I am right now.

PierUgit:
@everythingfunctional Thank for the proposal. I've also sent a message on the J3 ML in case someone would want to push it. Let's see.

certik:
My suggestion as the first step would be to get the committee to agree that this is in scope and should move forward, i.e. that you get support for this in the appropriate subgroup at least. For that 1. is helpful. @PierUgit once you secure the preliminary approval, then it's time to do 2. - 5. And as @everythingfunctional  said, it would help if you could agree that you will do it, if the committee gives a "preapproval".

everythingfunctional:
@certik , yes. We are currently at the phase of deciding what goes on the F202Y worklist. WG5 votes on the final list at the meeting in June. @PierUgit , you should view this paper as simply "I think the committee should work on this, and here's some initial ideas". It is basically pre step 1. But a tacit agreement that you will help with that work can make a difference in whether the committee will accept it.

PierUgit:
With some help I can definitely try.

PierUgit:
I have been thinking again about the alternative memory layouts in the 4.1 paragraph.

An hypothetical future row major storage isn't an issue at all actually. I imagine that it would be introduced through a new attribute in the declarations, e.g.
`real, layout(rowmajor), allocatable, target :: r(:,:)`
Then in `c => r` (or whatever retained syntax), `c` shall simply have the same layout:
`complex, layout(rowmajor), pointer :: c(:,:)`

Regarding the split-complex layout, I wonder if the proposal make any sense with it. The initial motivation is entirely linked to the in-place real to complex FFTs, and I struggle finding other usages. To me it's enough to justify the proposal, but I can't see how an in-place real to complex FFT would work with the split-complex layout. It would be a bit messy actually... So, technically the proposal may work for this layout, as described in the text, but IMO it would be pointless and without practical benefit.

I'm inclined to completely drop the part on the split-complex layout. If it is introduced in the future, the pointer association could simply be disallowed with this layout.

everythingfunctional:
@PierUgit , did you still want me to submit this paper for the meeting in two weeks?

PierUgit:
@everythingfunctional Definitely, if you agree...

everythingfunctional:
Ok, I will submit it and let you know about follow-up if any is needed.

everythingfunctional:
It's submitted: https://j3-fortran.org/doc/year/24/24-129.txt

PierUgit:
Great, thanks!

everythingfunctional:
@PierUgit , you'll be happy to learn that J3 did find this idea interesting enough to pursue for F202Y, and WG5 is highly likely to put it on the work list this morning.

PierUgit:
That's good, and thanks for the help! BTW, now that the draft has been officially submitted to the meeting, the branch could be merged (I have updated the document by using the uploaded version).

certik:
@PierUgit thanks for the proposal. I merged it, as requested. You can open a new PR for a subsequent paper(s).

@everythingfunctional thanks for representing the paper at the committee. This is a great example how the committee can involve the very large community out there to help. It works.

PR #364, opened 2026-02-20
raw draft for us11, source location routines

LorriMenard:


proposals/26-SourceLocation.txt
```diff
@@ -0,0 +1,169 @@
...
+In the case of a MODULE procedure the name will be the original name 
+from the module (not a possible rename).
+In the case of an instantiated TEMPLATE procedure, the name will be the
+ one given by the user.
+In the case this is used to create a PARAMETER constant in the specifications
+section it will return the name of the program unit.  
+
+If the program unit is an unnamed main program, it will return the string "MAIN PROGRAM"
+
+****Straw vote:  Is there a better value?
```
sblionel: I think MAIN PROGRAM works.

sblionel (approved):
See my comments (and one change), but otherwise looks good. (I am not very familiar with github so don't know if saying Approve includes my changes.)

proposals/26-SourceLocation.txt
```diff
@@ -0,0 +1,169 @@
...
+UNKNOWN INPUT
+Something else?
+
+
+SOURCE_LINE()
+Description. Returns the current source line
+Class. Transformational function
+Argument. None
+
+Result characteristics. Default integer value of the current source line.
```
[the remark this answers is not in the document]
bonachea: Please, no.

The value already needs to be processor-dependent (at least until the preprocessor joins the standard), and a processor-dependent character just opens way too many additional bizarre options (e.g. what is the character set? does it use scientific notation? does it separate digits with a locale-dependent separator like `,`? does it contain some leading blanks? etc...). 

As you say internal I/O can be used to convert either way, but printing an `integer` into a string format of the user's choosing is much simpler and more robust than portably parsing an arbitrary unspecified character representation of a number back into an integer.

Additionally, specifying it as an `integer` means IIUC it could potentially be used to easily initialize an `integer, parameter`, for example as a deferred integer constant in a template instantiation. Parsing an character representation back into an integer within the initializer for an integer parameter seems much more difficult.

proposals/26-SourceLocation.txt
```diff
@@ -0,0 +1,169 @@
+To: J3                                                     J3/26-xxx
+From: Lorri Menard & JoR
+Subject: US11 - Provide intrinsics for source location
```
bonachea: Suggest mentioning this is a specs paper in the title.
```suggestion
Subject: US11: Specifications for intrinsics for source location
```

proposals/26-SourceLocation.txt
```diff
@@ -0,0 +1,169 @@
...
+While a developer could readily produce such a string, that string will 
+often become incorrect as code undergoes other refactoring.  
+Further, experience shows that developers often use cut-and-paste when 
+developing new procedures ending up with wildly wrong names being passed 
+to the diagnostic layer.
+
+A preprocessor directive in a TEMPLATE declaration will be resolved at
+compile-time of the TEMPLATE.  However, instantiation of that template
+will happen in other files at other lines.  For debugging or logging
+purposes the value of __FILE__ and/or __LINE__ would not be helpful.
```
bonachea: Suggest we soften this wording, to avoid making assertions about a what's needed in a hypothetical debugging or logging scenario.

```suggestion
purposes the value of __FILE__ and/or __LINE__ might not be helpful.
```

proposals/26-SourceLocation.txt
```diff
@@ -0,0 +1,169 @@
...
+  CONTEXT(optional)   Shall be of type logical.  It is an INTENT(IN) argument.
+Result Characteristics. 
+The result is a default character scalar with processor-dependent length.
+
+Result Value.
+If CONTEXT is not present or has the value .FALSE. the processor shall return
+a filesystem name of the source input.
+If CONTEXT is present and has the value .TRUE. the processor may amend that
+string with more detailed filesystem information.
+If no information is available the value is the string "UNKNOWN FILE"
```
bonachea: I realize this is specifications, not edits, but the Fortran standard does not include a concept of "filesystem" (nor does it even acknowledge that program units are associated with a "file").

In our meeting I think we agreed to couch this all as processor-dependent, perhaps something like:

```suggestion
If CONTEXT is absent or has the value .FALSE., the result shall be
a processor-dependent value representing the name of the source input.
If CONTEXT is present with the value .TRUE., the result shall be a 
a processor-dependent value representing a potentially more detailed 
name of the source input.
If no information is available the value is the string "UNKNOWN FILE"

Example.
SOURCE_FILE() might have the value 'program.F90' and SOURCE_FILE(.true.) might have the value '/usr/src/program.F90'.
```

proposals/26-SourceLocation.txt
```diff
@@ -0,0 +1,169 @@
...
+UNKNOWN INPUT
+Something else?
+
+
+SOURCE_LINE()
+Description. Returns the current source line
+Class. Transformational function
+Argument. None
+
+Result characteristics. Default integer value of the current source line.
```
bonachea: I realize this is not edits, but "Result Characteristics." is strictly type information and should not record semantics.


```suggestion
Result characteristics. Default integer.
```

proposals/26-SourceLocation.txt
```diff
@@ -0,0 +1,169 @@
...
+By default this is computed from the start of the file where the first
+line is counted as 1.  If a #line preprocessor directive is seen that
+will establish a new source line number.
+It is expected that SOURCE_LINE() and the preprocessor command __LINE__
+return the same value.
+
+
+PROCEDURE_NAME([CONTEXT])
+Description. Processor-dependent string that describes the name of the
+             current procedure being executed.
```
bonachea: This description says "Processor-dependent", but the semantics below seem pretty rigidly specified. What exactly remains "processor-dependent" here? Capitalization?

bonachea (changes requested):
This is a great start!

I've noted a number of comments and suggestions.

proposals/26-SourceLocation.txt
```diff
@@ -0,0 +1,169 @@
...
+Argument. None
+
+Result characteristics. Default integer value of the current source line.
+
+Result value.
+By default this is computed from the start of the file where the first
+line is counted as 1.  If a #line preprocessor directive is seen that
+will establish a new source line number.
+It is expected that SOURCE_LINE() and the preprocessor command __LINE__
+return the same value.
```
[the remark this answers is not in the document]
LorriMenard: Not the program unit really; a file with multiple external subroutines has multiple program units - the number wouldn't necessarily restart at the top of each of them.  Going to simply use "program"

proposals/26-SourceLocation.txt
LorriMenard: Or mangled name

proposals/26-SourceLocation.txt
```diff
@@ -0,0 +1,181 @@
...
+contains
+   subroutine s()
+      print*, procedure_name()       ! 'S' 
+      print*, procedure_name(.TRUE.) ! 'A(B):S' or '(A)B:S'or 'A/B:S'
+   contains
+      subroutine inner()
+         print*, procedure_name()       ! 'inner'
+         print*, procedure_name(.TRUE.) ! 'A(B):s.inner' or
+                                        ! '(A)B:s.inner' or
+                                        ! 'A/B:S.inner'
```
bonachea, line 174: Some of these examples preserve the name capitalization given in the source program, and others do not.

IMO we should either:

1. Require the return value preserve the capitalization given in the name as it appears in the appropriate `*-name` in the source program
2. Require the return value to always be upper-case.
3. Make the capitalization of the return value explicitly processor dependent.

I suggest a straw poll to settle this.

LorriMenard: I agree that I was inconsistent - I did briefly notice that, and forgot to go back.
I *think* the standard in examples is uppercase (except for binding name) but another straw poll is not a bad idea.

proposals/26-SourceLocation.txt
LorriMenard: I was thinking something along this line (wrt generic)

      GENERIC FUNCTION DIVIDE (X, Y)
       TYPE(INTEGER(2,4)) :: X
       TYPE(INTEGER(2,4)) :: Y
       IF (Y == 0)
            PRINT *, "Problem in ", PROCEDURE_NAME()
            ERROR STOP
       ENDIF
       DIVIDE = X / Y
      END FUNCTION DIVIDE

This is a braindead example, and this usage would be more compelling with other data types and error conditions.
What might be more interesting is as a trace function, to log which incarnations you actually used.

In any case, I'm not married to this thought.

bonachea:
Superseded by : 
* [26-129r1](https://j3-fortran.org/doc/year/26/26-129r1.txt) US11 - Edits for intrinsics for source location

PR #367, opened 2026-05-27
Edits to correct prefix reductions (US20)

bonachea:


bonachea:
Uploaded as https://j3-fortran.org/doc/year/26/26-159.txt

corrections/26-159r2.txt
```diff
@@ -0,0 +1,334 @@
...
+
+Despite the "General rule" from 17.2.1 p3, subclause 17.9 currently includes
+45 instances of "INTENT (IN)" appearing in the description of dummy arguments
+to various intrinsic subroutines.  Of all the dummy arguments to all the
+intrinsic subroutines appearing in subclause 17.9, exactly four dummy argument
+descriptions are missing an explicit "INTENT" in 26-007r1. Those are:
+
+* CO_REDUCE: OPERATION argument
+* CO_REDUCE_PREFIX_EXCLUSIVE: OPERATION and INITIAL arguments
+* CO_REDUCE_PREFIX_INCLUSIVE: OPERATION argument
```
everythingfunctional, line 67: OPERATION is a procedure argument, and procedure arguments do not have INTENT.

bonachea: Thanks for the feedback @everythingfunctional!

Well if you're correct I think that's a new problem because 17.2.1p3 reads:

> The dummy arguments of the intrinsic procedures in 17.9 have INTENT (IN) if the intent is not stated explicitly.

It does NOT say "unless they happen to be a dummy procedure" (although perhaps it should?)

In any case, technically there is no _actual_ Fortran declaration for `OPERATION`, the intrinsic merely states:

> OPERATION shall be a pure function with exactly two arguments;

So I'm not sure that C850 applies anyhow. For example, OPERATION could perhaps be _implemented_ as a dummy procedure pointer, which does allow `INTENT(IN)` on the declaration (although in my experience compilers don't generally declare interfaces for the generic intrinsics anyhow).  

My goal here was mostly to clarify that if the user _did_ pass a pointer to a pure function in OPERATION that meets all the other rules, the INTENT(IN) would guarantee that pointer is not modified by the subroutine invocation.

